PR #1, opened 2023-08-14
Restructuring repository

dutow:
* All current code moved to pg_tde subdirectory
* Makefile adjusted
* Meson build file added
* Gitignore added to ignore in-source files with make


EngineeredVirus:
The repository is restructured in a different way.

PR #3, opened 2023-08-17
Fix compilation with PGXS

dAdAbird:
```
make USE_PGXS=1
```
g̶r̶e̶a̶t̶ work again


dutow (approved):
You should also delete the last line in the Makefile (`$(recurse)`). Otherwise looks good.

PR #5, opened 2023-08-22
Code re-arrangement, Makefile fixes and implementing '.tde' relation fork

codeforall:
- Fixing the Makefile to link with ssl and crypto libs required by the extension.

- An entry file pg_tde.c for all extension-related initialization functions

- A a new ".tde" relation fork to store the encryption key and related data. Currently, the relation fork gets created with new relations and a dummy encryption key gets stored in the fork.

- Re-arranging the directory structure to separate source and header files


Makefile
```diff
@@ -1,23 +1,39 @@
 # contrib/pg_tde/Makefile
 
 PGFILEDESC = "pg_tde access method"
-
+MODULE_big = pg_tde
```
dutow, line 4: Could you please also adjust the meson buildfile, so both builds keep working?

codeforall: Pushed the fix as discussed over slack

dAdAbird (commented):
Here https://github.com/Percona-Lab/postgres-tde-ext/pull/5/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R17
should be 
```
incdir = include_directories('src/include')
```

dutow (approved):
Looks good (that remaining CI failure is unrelated)

dAdAbird (approved):

PR #6, opened 2023-08-22
Use postgres mem context in decryption

dAdAbird:


src/encryption/enc_tuple.c
```diff
@@ -79,8 +80,12 @@ static void PGTdeDecryptTupInternal2(BlockNumber bn, Page page, HeapTuple tuple,
 	// Most of the time we can't decrypt in place, so we allocate some memory... and leek it for now :(
```
dutow, line 80: This comment is no longer true with these changes.

dAdAbird: yep, I even checked if there is no comment about the leak - thought it was right above malloc :)
fixed https://github.com/Percona-Lab/postgres-tde-ext/pull/6/commits/0738044787e8af9a468d2d9303a1577e86427ff6

dutow (approved):

codeforall (approved):

PR #8, opened 2023-08-23
Minimal keyring prototype

dutow:
This commit implements a very basic file based keyring, intended for development.

To use the server after these changes, the following initialization steps are required:

* pg_tde should be preloaded: ALTER SYSTEM SET shared_preload_libraries = 'pg_tde';
* pg_tde extension should be installed
* pg_tde.keyringConfigFile = '<path>' should be set in postgresql.conf or command line
* The config file is a simple json file
* After the variable is set, and the config file exists, the server needs a restart

Example config file:

{
        'provider': 'file',
        'datafile': '/tmp/pgkeyring.data',
}


src/Makefile
```diff
@@ -0,0 +1,19 @@
+# contrib/pg_tde/Makefile
```
dAdAbird, line 1: I don't think we need this file anymore. It was removed in https://github.com/Percona-Lab/postgres-tde-ext/commit/9fa73a60874efaadb8acd9f958ff7ce0e4a298bd#diff-3e2513390df543315686d7c85bd901ca9256268970032298815d2f893a9f0685

dAdAbird (approved):

PR #10, opened 2023-08-28
Adding infrastructure to clean files based on transaction status

codeforall:
.tde fork creation is also made transactional.

### DIFF
--- a/src/transam/pg_tde_xact_handler.c
+++ b/src/transam/pg_tde_xact_handler.c
@@ -1,0 +1,187 @@
+/*-------------------------------------------------------------------------
+ *
+ * pg_tde_xact_handler.c
+ *	  Transaction handling routines for pg_tde
+ *
+ *
+ * IDENTIFICATION
+ *	  src/transam/pg_tde_xact_handler.c
+ *
+ *-------------------------------------------------------------------------
+ */
+
+#include "postgres.h"
+#include "utils/memutils.h"
+#include "utils/palloc.h"
+#include "utils/elog.h"
+#include "storage/fd.h"
+#include "transam/pg_tde_xact_handler.h"
+
+typedef struct PendingFileDelete
+{
+    char    *path;    /* file that may need to be deleted */
+    bool    atCommit;       /* T=delete at commit; F=delete at abort */
+    int     nestLevel;      /* xact nesting level of request */
+    struct PendingFileDelete *next;  /* linked-list link */
+} PendingFileDelete;
+
+static PendingFileDelete *pendingDeletes = NULL; /* head of linked list */
+
+static void cleanup_pending_deletes(bool atCommit);
+static void do_pending_deletes(bool isCommit);
+static void pending_delete_cleanup(void);
+
+/* Transaction Callbacks from Backend*/
+void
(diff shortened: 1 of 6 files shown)


src/transam/pg_tde_xact_handler.c
```diff
@@ -0,0 +1,187 @@
...
+{
+    char    *path;    /* file that may need to be deleted */
+    bool    atCommit;       /* T=delete at commit; F=delete at abort */
+    int     nestLevel;      /* xact nesting level of request */
+    struct PendingFileDelete *next;  /* linked-list link */
+} PendingFileDelete;
+
+static PendingFileDelete *pendingDeletes = NULL; /* head of linked list */
+
+static void cleanup_pending_deletes(bool atCommit);
```
dutow, line 30: These function are using a different_naming_convention ThanTheOthers.

src/transam/pg_tde_xact_handler.c
```diff
@@ -0,0 +1,187 @@
...
+static void do_pending_deletes(bool isCommit);
```
dutow, line 31: Do we really need the bool parameter for this function and the previous? It seems we are only using one value for them.

src/transam/pg_tde_xact_handler.c
```diff
@@ -0,0 +1,187 @@
...
+
+void
+RegisterFileForDeletion(const char *filePath, bool atCommit)
+{
+    PendingFileDelete *pending;
+    pending = (PendingFileDelete *) MemoryContextAlloc(TopMemoryContext, sizeof(PendingFileDelete));
+    pending->path = MemoryContextStrdup(TopMemoryContext, filePath);
+    pending->atCommit = atCommit;  /* delete if abort */
+    pending->nestLevel = GetCurrentTransactionNestLevel();
+    pending->next = pendingDeletes;
```
dutow, line 78: Shouldn't this list have a lock?

src/transam/pg_tde_xact_handler.c
codeforall: Yes, that is true. Right now, we only use the functions to delete files at transaction Abort. But since I was creating the infrastructure for making the files transactional aware, I added this bool so that in the future, when we want to clean up some temporary file at transaction commit, we should already have the infrastructure for that.

src/transam/pg_tde_xact_handler.c
codeforall: I don't think so, Since every backend is maintaining this list in its local space, there can't be any contention on the list

src/transam/pg_tde_xact_handler.c
codeforall: The convention I followed is to name static functions as a_b_c() while the global ones as ThisOne(). PG is not 100% consistent everywhere in the function naming convention and uses the following in some places.  For callback functions pg_tde_subxact_callback and pg_tde_xact_callback, though they are not static in the file everywhere, they uses a_b_c() naming convention, so I stuck with the same. But I agree your observation is correct, and we may need to document and follow one particular naming convention throughout the code. If it's Okay we can differ this comment for future ?

dutow (approved):

dAdAbird (approved):

PR #12, opened 2023-08-31
Store and fetch relation keys from fork

dAdAbird:


src/include/access/pg_tde_tdemap.h
```diff
@@ -11,6 +11,47 @@
...
+{
+    uint8   key[INTERNAL_KEY_LEN];
+    /* start and end range of the key
+     * (start_loc == 0 && end_loc == 0) -> the key if for the whole file
+     */
+    Size    start_loc; 
+    Size    end_loc;
+} InternalKey;
+
+#define MASTER_KEY_LEN 256
```
dutow, line 27: Maybe MASTER_KEY_NAME_LEN instead?

src/access/pg_tde_tdemap.c
```diff
@@ -10,49 +10,188 @@
...
+		ereport(FATAL,
+                (errcode_for_file_access(),
+                errmsg("could not open tde key file \"%s\": %m",
+						key_file_path)));
+
+	
+	sz = (Size) FileSize(file);
+	keys = (RelKeysData *) MemoryContextAlloc(TopMemoryContext, sz);
+
+	nbytes = FileRead(file, keys, sz, 0, WAIT_EVENT_DATA_FILE_READ);
```
dutow, line 114: We need at least a todo comment here, and where we write the file, that the internal keys should be encrypted in the future.

src/include/access/pg_tde_tdemap.h
dAdAbird: https://github.com/Percona-Lab/postgres-tde-ext/pull/12/commits/56abaac403faa89142e656516e902537c71e3b5d

src/access/pg_tde_tdemap.c
dAdAbird: https://github.com/Percona-Lab/postgres-tde-ext/pull/12/commits/56abaac403faa89142e656516e902537c71e3b5d

dutow (approved):

codeforall (approved):

PR #14, opened 2023-09-07
Fix crash in tde files: next was possibly uninitalized

dutow:
Also fixes a warning related the possible uninitialized use of prev.


codeforall:
LGTM

PR #16, opened 2023-09-11
Deleting respective tde fork file with Drop Table

codeforall:


dAdAbird (approved):

src/access/pg_tde_tdemap.c
```diff
@@ -24,6 +24,19 @@
 #include <openssl/rand.h>
 #include <openssl/err.h>
 
+void
+pg_tde_delete_key_fork(Relation rel)
+{
+	char		*rel_file_path;
+	char		*key_file_path;
+
+	rel_file_path = relpathperm(rel->rd_locator, MAIN_FORKNUM);
```
dutow, line 33: We could introduce a helper function this instead of doing the same logic in multiple functions.

dutow (approved):
Looks good, only added a comment about a possible minor improvement.

dutow:
@codeforall could you please rebase on the now force-pushed tuple_enc branch?

PR #21, opened 2023-09-15
Fix compaction of non-presorted tuples

dAdAbird:
That covers the case when line pointers (lp) point to unsorted offsets (tuples):
```
SELECT lp, lp_off, t_ctid FROM heap_page_items(get_raw_page('sbtest1', 0));
 lp | lp_off | t_ctid
----+--------+--------
  1 |   7960 | (0,1)
  2 |   7264 | (0,2)
  3 |   7032 | (0,3)
  4 |   6800 | (0,4)
  5 |   7728 | (0,5)
  6 |      0 |
  7 |   7496 | (0,7)
(7 rows)
```

This condition can be achieved by deleting some tuples and running VACUUM

Also, the encoded bytes printing is hidden behind higher ENCRYPTION_DEBUG. It floods log a lot otherwise.

### DIFF
--- a/src/access/pg_tde_prune.c
+++ b/src/access/pg_tde_prune.c
@@ -1417,7 +1417,6 @@ pgtde_compactify_tuples(Relation rel, Buffer buffer, itemIdCompact itemidbase, i
 	}
 	else
 	{
-		Assert(0); // TODO: to be implemented with encryption
 		PGAlignedBlock scratch;
 		char	   *scratchptr = scratch.data;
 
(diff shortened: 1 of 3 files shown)


dutow:
> This condition can be achieved by deleting some tuples and running VACUUM

Can you add an SQL script doing exactly this (and crashing without the fix) to the repository somewhere, for example in a new `test` directory? That would help us writing actual testcases for CI later. (I also forgot this for the previous crash fix, I'll create a separate PR for that)

dutow (approved):

PR #30, opened 2023-09-26
Fix indexes comparison on UPDATE

dAdAbird:
There is a check during the UPDATE for changed indexes. Sometimes it causes a crash since the new tuple is unencrypted and the old is encrypted.
Besides comparison didn't work properly. Which leads to taking a suboptimal code path. And probably computing wrong replica identity and freeing used tuple in some cases.


codeforall (approved):
Fixes the primary issue mentioned in the PR.

PR #42, opened 2023-10-10
Basic is encrypted check function

### LINKED ISSUE #40: Add a function to validate if my data is encrypted
As of today, the only way to check if my data was encrypted is to check the access method, and if pg_tde was used, it is encrypted.

This is complicated, and not user-friendly. We should add a function, that easily returns and shows if my data is encrypted or not.

dutow:
This commit adds a new SQL function, `pgtde_is_encrypted(tablename)` which returns a boolean value: true if the table is encrypted with pg_tde, false otherwise.

As the pg_tde access method only writes encrypted data, it just checks if the table uses the am.

Fixes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,14 @@ You can also build a docker image manually with:
 docker build . -f ./docker/Dockerfile -t your-image-name
 ```
 
+## Helper functions
+
+The extension provides the following helper functions:
+
+### pgtde_is_encrypted(tablename)
+
+Returns `t` if the table is encrypted (uses the pg_tde access method), or `f` otherwise.
+
 ## Base commit
 
 This is based on the heap code as of the following commit:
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -111,6 +111,14 @@ You can also build a docker image manually with:
 docker build . -f ./docker/Dockerfile -t your-image-name
 ```
 
+## Helper functions
+
+The extension provides the following helper functions:
+
+### pgtde_is_encrypted(tablename)
```
capri1989, line 118: How do you call the helper function? Can you simply use the function within psql shell?

dutow: Yes, all SQL functions can be called in the form of `SELECT pgtde_is_encrypted(foo)`.

PR #47, opened 2023-10-16
Encrypt fork files with master key and improve tests

### LINKED ISSUE #46: No errors, even if user don't sepcify any Keyring configuration
```
postgres=# show pg_tde.keyringconfigfile ;
 pg_tde.keyringConfigFile 
--------------------------
 
(1 row)

postgres=# \dt+
                                      List of relations
 Schema |  Name   | Type  |  Owner   | Persistence | Access method |    Size    | Description 
--------+---------+-------+----------+-------------+---------------+------------+-------------
 public | sbtest1 | table | postgres | permanent   | pg_tde        | 0 bytes    | 
 public | t1      | table | postgres | permanent   | heap          | 8192 bytes | 
 public | t2      | table | postgres | permanent   | pg_tde        | 8192 bytes | 
 public | t3      | table | postgres | permanent   | heap          | 8192 bytes | 
(4 rows)

postgres=# select * from t2;
 id 
----
  1
  2
(2 rows)

postgres=# create table t4 (id int) using pg_tde;
CREATE TABLE
```

dutow:
Fork file encryption was missing from the previous pull requests, resulting in the server not initializing the keyring on normal executions, and because of this, missing if the keyring configuration wasn't specified at all.

Fixes #46


src/encryption/enc_aes.c
```diff
@@ -37,23 +38,24 @@
...
 
 void AesInit(void)
 {
 	static int initialized = 0;
 
 	if(!initialized) {
-		// Not sure if we need this 2, does postgres already call them somewhere?
+
+		fprintf(stderr, "Initializing!\n");
+
```
dAdAbird, line 52: I suppose this is a dev leftovers and can be removed

dAdAbird (approved):

PR #50, opened 2023-10-19
code cleanup and re-arrangements for tuple encryption and decryption

codeforall:


codeforall:
The PR needs a few more fixes

PR #52, opened 2023-10-20
Few enhancements and code cleanup around tuple encryption/decryption

codeforall:
Along with code cleanup, the PR contains the following noteworthy changes
-- Getting rid of VLAs from the code base
-- Add an interface to move the encrypted data from one location to another.
-- Make the encryption and decryption happen in batches to eliminate the requirement of dynamic allocation in crypt functions.


dAdAbird (approved):

src/encryption/enc_tuple.c
```diff
@@ -9,7 +9,14 @@
 #include "storage/bufmgr.h"
 #include "keyring/keyring_api.h"
 
-#define AES_BLOCK_SIZE 16
+#ifdef ENCRYPTION_DEBUG
+	/* While in active development, We are emmiting a LOG message for debug data when ENCRYPTION_DEBUG is enabled.*/
+	const int enc_log_elevel = LOG;
+#else
+	const int enc_log_elevel = DEBUG2;
```
dutow, line 16: This is a bad idea: the reason why we don't log anything when ENCRYPTION_DEBUG is off is because of the huge performance hit. Even if the log level is DEBUG2, and it's not enabled, the message will have to go through the logging system logic.

With these changes, the performance hit of encyption drops from <10% to 50%+, and that's with a release build and the debug macros disabled.

We can use the log system if we want to instead of fprintf (not that it makes much difference for these development-only high output debug prints), but please make sure that if ENCRYPTION_DEBUG/KEYRING_DEBUG/... is not defined, we don't call anything related to logging.

dutow (changes requested):
Partial review - I only ran benchmarks because I was curious if the VLA changes cause any measurable change.

Please see my comment, I found a different performance issue in this.

src/encryption/enc_tuple.c
codeforall: PR Updated ,with ereport removed when ENCRYPTION_DEBUG is not defined.

src/encryption/enc_tuple.c
```diff
@@ -44,9 +37,11 @@ pg_tde_crypt(uint64 start_offset, const char* data, uint32 data_len, char* out,
 		uint64 batch_end_block = Min(batch_start_block + NUM_AES_BLOCKS_IN_BATCH, aes_end_block);
 
 	    Aes128EncryptedZeroBlocks2(&(keys->internal_key[0].ctx), keys->internal_key[0].key, batch_start_block, batch_end_block, encKey);
-		ereport(enc_log_elevel,
+#ifdef ENCRYPTION_DEBUG
```
dutow, line 40: Please keep the old `#if ENCRYPTION_DEBUG` style. `#ifdef` evaulates to true with `#define ENCRYPTION_DEBUG 0`, `#if` evaluates to false.

src/include/encryption/enc_aes.h
```diff
@@ -1,15 +1,29 @@
...
+#ifndef ENC_AES_H
+#define ENC_AES_H
 
 #include <stdint.h>
 
-void AesInit(void);
+#define AES_BLOCK_SIZE 		        16
+#define NUM_AES_BLOCKS_IN_BATCH     100
+#define DATA_BYTES_PER_AES_BATCH    (NUM_AES_BLOCKS_IN_BATCH * AES_BLOCK_SIZE)
+#define MAX_AES_ENC_KEY_SIZE        ((NUM_AES_BLOCKS_IN_BATCH + 1) * AES_BLOCK_SIZE)
```
dutow, line 18: I would use a different name for this - this isn't the size of the key, but the maximum data size when encrypting/decrypting in batches.

src/encryption/enc_aes.c
```diff
@@ -113,10 +113,7 @@ void Aes128EncryptedZeroBlocks(const unsigned char* key, uint64_t blockNumber1,
 	unsigned char iv[16] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00 };
 
 	unsigned dataLen = (blockNumber2 - blockNumber1 + 1) * 16;
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wvla"
-	unsigned char data[dataLen];
-#pragma GCC diagnostic pop
+	unsigned char data[MAX_AES_ENC_KEY_SIZE];
```
dutow, line 116: Please add at least an assert that dataLen is smaller.

src/encryption/enc_aes.c
```diff
@@ -147,10 +144,7 @@ void Aes128EncryptedZeroBlocks2(void* ctxPtr, const unsigned char* key, uint64_t
...
```
dutow, line 147: Please add at least an assert that datalen is smaller.

src/encryption/enc_tuple.c
```diff
@@ -21,97 +21,123 @@
...
-    encKey = palloc(AES_BLOCK_SIZE * (aes_end_block - aes_start_block + 1));
-
-    // TODO: verify key length!
-    Aes128EncryptedZeroBlocks2(&(keys->internal_key[0].ctx), keys->internal_key[0].key, aes_start_block, aes_end_block, encKey);
-
-#if ENCRYPTION_DEBUG
-    /* While in active development, We are emmiting a LOG message for debug data when ENCRYPTION_DEBUG is enabled.*/
-    ereport(LOG,
-        (errmsg("%s: Start offset: %llu Data_Len: %u, AesBlock: %llu, BlockOffset: %llu", context?context:"", start_offset, data_len, aes_start_block, aes_block_no)));
+    uint64 aes_start_block = start_offset / AES_BLOCK_SIZE;
```
dutow, line 26: Why do we have to change this in a code cleanup? We are using C99, where `uint64_t` is the standard type for 64 bit unsigned integers, defined in `stdint.h`. `uint64` is a non standard type.

Also, these variables shouldn't change once calculated, so declaring them `const` is a good practice.

src/encryption/enc_tuple.c
```diff
@@ -21,97 +21,123 @@
...
+    uint64 aes_end_block = (start_offset + data_len + (AES_BLOCK_SIZE -1)) / AES_BLOCK_SIZE;
+    uint64 aes_block_no = start_offset % AES_BLOCK_SIZE;
+	unsigned char encKey[MAX_AES_ENC_KEY_SIZE];
+	uint64 block_no;
+	uint32 batch_no = 0;
+
+	/* do max NUM_AES_BLOCKS_IN_BATCH blocks at a time */
+	for (block_no = aes_start_block; block_no < aes_end_block; block_no += NUM_AES_BLOCKS_IN_BATCH)
```
dutow, line 34: `for (uint64_t block_no = ...)`?

src/encryption/enc_tuple.c
```diff
@@ -21,97 +21,123 @@
...
-    pfree(encKey);
+		for(uint32 i = 0; i < DATA_BYTES_PER_AES_BATCH; ++i)
+		{
+			uint32 data_index = i + (batch_no * DATA_BYTES_PER_AES_BATCH);
+			uint32	enc_key_index = i + (batch_no > 0 ? 0 : aes_block_no);
+			if (data_index >= data_len)
+				break;
+
+			#if ENCRYPTION_DEBUG > 1
+	        fprintf(stderr, " >> 0x%02hhX 0x%02hhX\n", v & 0xFF, (v ^ encKey[aes_block_no + i]) & 0xFF);
```
dutow, line 54: This is still an `fprintf`, is this intentional?

src/encryption/enc_tuple.c
codeforall: It's entirely a question of style. And IMHO, for enabling/disabling debugging code, it's always easier to spot and use #ifdef regardless of its value.  I am also inclined towards using #ifdef style since it is more consistent with our borrowed code styling. e.g. #ifdef USE_ASSERT_CHECKING and #ifdef USE_PREFETCH

src/include/encryption/enc_aes.h
codeforall: What do you think about MAX_BATCH_AES_ENC_KEY_SIZE? or what is your suggestion?

src/encryption/enc_aes.c
codeforall: will do

src/encryption/enc_aes.c
codeforall: will do

src/encryption/enc_tuple.c
codeforall: uint64 is the standard across PostgreSQL code. So, it's better to stick with PG conventions.

src/include/encryption/enc_aes.h
codeforall: Changed to MAX_AES_ENC_BATCH_KEY_SIZE

codeforall:
All review comments have been resolved, so merging the PR.

PR #55, opened 2023-10-23
Basic performance tests on github.

dutow:
This also adds a slighly modified sysbench to the repository, which creates encrypted tables, and modifies the debug message default to 0.


dutow:
`workflow_run` triggers only works on the main branch, and because of that, the comment is not visible here. This is an example pull request with everything working: https://github.com/dutow/postgres-tde-ext/pull/1/

dAdAbird (approved):

codeforall (approved):

PR #67, opened 2023-11-08
Fix update crash after code cleanup

dutow:
Issue: the code cleanup introduced in PR #52 modified the original tuple in the update method instead of decrypting the tuple data into a copy. This caused data curruption crashes in some test.

Fix: reintroduce the missing palloc to the update method.

Fixes #61
Fixes #62
Fixes #64

PR #69, opened 2023-11-13
Fix update duplication after crashfix

### LINKED ISSUE #68: UPDATE statement broken after PR67
After [67](https://github.com/Percona-Lab/postgres-tde-ext/pull/67), UPDATE statement adds new rows instead of updating the previous ones 
```

DROP TABLE IF EXISTS test_update;
CREATE TABLE test_update ( a int) using pg_tde;
INSERT INTO test_update VALUES (1), (2),(3), (4);
SELECT * from test_update;
UPDATE test_update SET a = a + 1;
--Expecting 4 rows
SELECT * from test_update;
 a 
---
 1
 2
 3
 4
 2
 3
 4
 5
(8 rows)

```

dutow:
Issue: with the fix in #67, pgtde decrypts tuples during update into a new memory region, and changes the t_data pointer to this new region.

Because of this, later updates to tuple flags also happen in the new data, and the original persisted tuple flags are never updated.

Fix: after the update statement is done with the decrypted data, restore the t_data pointer to the original. This way, flag changes happen where they should.

Fixes #68

### DIFF
--- a/expected/trigger_on_view.out
+++ b/expected/trigger_on_view.out
@@ -157,8 +157,7 @@ CONTEXT:  PL/pgSQL function city_update() line 9 at RAISE
  city_id | city_name | population | country_name | continent 
 ---------+-----------+------------+--------------+-----------
        1 | Tokyo     |   13010279 | Japan        | Asia
-       1 | Tokyo     |   13010279 |              | 
-(2 rows)
+(1 row)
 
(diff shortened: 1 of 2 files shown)


dAdAbird (approved):

codeforall (approved):
Looks perfect

Issue #72, opened 2023-11-15
Broken encryption / decryption of 2+Kb data


dAdAbird:
1. TOASTed data bigger around 5Kb and more being broken on SELECT.
2. Some tuples of the size 2.5Kb being broken after moving (compaction of the pages).

PR #73, opened 2023-11-15
Fix decryption of large tuples/TOASTs

dAdAbird:
Encryption/decryption of the same data should be exact as long as the offset is the same. But as we encode in 16-byte blocks, the size of `encKey` always is a multiple of 16. We start from `aes_block_no`-th index of the encKey[] so N-th will be crypted with the same encKey byte despite what start_offset `pg_tde_crypt()` was called with. For example `start_offset = 10; MAX_AES_ENC_BATCH_KEY_SIZE = 6`:
```
data:                 [10 11 12 13 14 15 16]
encKeys: [...][0 1 2 3  4  5][0  1  2  3  4  5]
```
so the 10th data byte is encoded with the 4th byte of the 2nd encKey
etc. We need this shift so each byte will be coded the same despite the
initial offset.
Let's see the same data but sent to the func starting from the offset 0:
```
data:    [0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16]
encKeys: [0 1 2 3 4 5][0 1 2 3 4 5][ 0 1  2  3  4  5]
```
again, the 10th data byte is encoded with the 4th byte of the 2nd
`encKey` etc.

The issue in `pg_tde_crypt()` was that along with shifting encKeys in the first batch, we started skipping `aes_block_no` bytes in `data` with every batch after the first one. That lead to:
1. Some bytes remained unencrypted.
2. If data was encrypted and decrypted in different batches these skipped bytes would differ hence decrypted data would be wrong. TOASTed data, for example, is encrypted as one chunk but decrypted in `TOAST_MAX_CHUNK_SIZE` chunks.

The issue with pg_tde_move_encrypted_data() was that encryption and decryption were happening in the same loop but `encKeys` may have had different shifting as in and out data may have different start_offset. It wasn't an issue if the data was less than `DATA_BYTES_PER_AES_BATCH`. However, it resulted in data corruption while moving the larger tuples. Plus, it makes sense to reuse `pg_tde_crypt()` instead of copying its tricky code.

Also, encKey had a maximum size of 1632 bytes but only 1600  bytes maximum could have been used. We don't need that extra buffer 32 bytes buffer anymore.

The same with the `dataLen` in `Aes128EncryptedZeroBlocks2()` - I don't see why do we need that extra block.

Fixes https://github.com/Percona-Lab/postgres-tde-ext/issues/72


src/encryption/enc_tuple.c
```diff
@@ -62,43 +87,16 @@ pg_tde_move_encrypted_data(uint64 read_start_offset, const char* read_data,
...
-			char decrypted_byte;
-
-			if (data_index >= data_len)
-				break;
-	        decrypted_byte = read_data[data_index] ^ read_encKey[read_enc_key_index];
-			write_data[data_index] = decrypted_byte ^ write_encKey[write_enc_key_index];
-		}
-		batch_no++;
+		pg_tde_crypt(read_start_offset + i, read_data + i, batch_len, decrypted, keys, context);
+		pg_tde_crypt(write_start_offset + i, decrypted, batch_len, write_data + i, keys, context);
```
codeforall, line 99: Now, we are decrypting and encrypting in separate cycles, while previously, it was done in a single loop. I am unsure, how often this function hits in a normal workload, but we can still do decryption and re-encryption in a single loop using the previous technique. One drawback of that would be that we would be effectively duplicating a lot of logic in this function, but if we can gain some performance out of it then it could be a good compromise. (Just a suggestion)

src/encryption/enc_tuple.c
```diff
@@ -11,48 +11,73 @@
...
  * ================================================================
- *
- * data and out have to be different addresses without overlap!
- * start_offset: is the absolute location of start of data in the file
- * The only difference between enc and dec is how we calculate offsetInPage
  */
 
+/* 
+ * pg_tde_crypt
+ * data and out have to be different addresses without overlap!
```
codeforall, line 22: I believe the function can do encryption/decryption in place, and we can remove this particular comment.

src/encryption/enc_tuple.c
dutow: I think the compiler should be able to optimize this loop the same way, but we could try to check a simplified version of both versions on godbolt

src/encryption/enc_tuple.c
```diff
@@ -11,48 +11,73 @@
...
-    uint64 aes_block_no = start_offset % AES_BLOCK_SIZE;
-	unsigned char encKey[MAX_AES_ENC_BATCH_KEY_SIZE];
+	uint64 aes_start_block = start_offset / AES_BLOCK_SIZE;
+	uint64 aes_end_block = (start_offset + data_len + (AES_BLOCK_SIZE -1)) / AES_BLOCK_SIZE;
+	uint64 aes_block_no = start_offset % AES_BLOCK_SIZE;
 	uint32 batch_no = 0;
+	uint32 data_index = 0;
+	uint64 batch_end_block;
+	uint32 current_batch_bytes;
+	unsigned char encKey[MAX_AES_ENC_BATCH_KEY_SIZE];
```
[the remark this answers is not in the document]
dAdAbird: Yes, I was thunking about abandoning MAX_AES_ENC_BATCH_KEY_SIZE but then forgot tbh. Will do

src/encryption/enc_tuple.c
dAdAbird: Keeping everything in one loop will complicate the things. As `in` and `out` could have different `start_offset` in relation to their `enc_key`s, means can't do full enc and dec in the same batch loop if the data_len more than `DATA_BYTES_PER_AES_BATCH` (`data_len` + `aes_block_no` to be precise) - we would have to keep reminder of the data with the bigger `aes_block_no` to the next iteration of the loop, encode (or decode) it with the next `enc_key` against the respective reminder of the data counterpart with its old enc_key and then proceed the rest with the all new `enc_key`s. Meaning a complicated code that I'd prefer to avoid for maintanence reasons. Of course, if it's give us a noticeable  performance advantage, I'm all in keep it in one loop. Let's do benchmarking then.  `pg_tde_move_encrypted_data()` being used only wile tuples moving/compaction. As far as I understand it happens only during VACUUMs if pages have some defragmentation.

dAdAbird:
Also removed an unused old version of Aes128EncryptedZeroBlocks() and made a couple of changes to get rid of compiler warnings

src/encryption/enc_tuple.c
```diff
@@ -80,23 +80,25 @@ pg_tde_crypt(uint64 start_offset, const char* data, uint32 data_len, char* out,
...
 */
 void
 pg_tde_move_encrypted_data(uint64 read_start_offset, const char* read_data,
 				uint64 write_start_offset, char* write_data,
 				uint32 data_len, RelKeysData* keys, const char* context)
 {
-	uint32 batch_len;
-	char decrypted[DATA_BYTES_PER_AES_BATCH];
+	uint32	batch_len;
+	uint32	byes_left;
```
dutow, line 91: Is this a typo (bytes_left)?

dAdAbird: oops, fixed

codeforall (approved):
LGTM

Issue #63, opened 2023-10-31
Crash/Data Corruption with TEMP tables containing toasted value


codeforall:
When a column of a temp table requires out-of-line storage. Inserting data in that temp table causes the data corruption/crash.

```
CREATE TEMP TABLE src (f1 text) using pg_tde;
INSERT INTO src              
   SELECT string_agg(random()::text,'') from generate_series(1,10000);
SELECT * from src;

```

dAdAbird:
Concerns non-TEMP tables as well

PR #76, opened 2023-11-20
Fix compressed TOASTs encryption/decryption

dAdAbird:
If TOAST data gets compressed, it has an extended header containing compression info. We used to encrypt this header along with the actual data which in turn caused a crash as PG needs this data in later stages. So it should be taken into account while encrypting data during externalisation.

Then, during detoasting, we should not decrypt this compression header as it is being extracted with the data with the first TOAST chunk. So, copy the first N bytes (now it is 4 bytes) of the first chunk as it is and decrypt the rest of the data.

Fixes https://github.com/Percona-Lab/postgres-tde-ext/issues/63


codeforall (approved):
LGTM.

PR #78, opened 2023-11-29
Docs

nastena1606:


documentation/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+Their encryption is planned for the next releases of `pg_tde`.
+
+<i warning>:material-alert: Warning:</i> Note that introducing encryption/decryption affects performance. Our benchmark tests show appr. 10% performance overhead.
+
+[Get started](install.md){.md-button}
+
+## Supported PostgreSQL versions
+
+`pg_tde` is currently supported for Percona Distribution for PostgreSQL 16 and PostgreSQL 16.
```
capri1989, line 34: For similarity with the installation guide, I would write "upstream PostgreSQL 16"

documentation/docs/test.md
```diff
@@ -0,0 +1,3 @@
+# Test Transparent Data Encryption
+
+To check if the data is encrypted, do the following
```
capri1989, line 3: https://github.com/Percona-Lab/postgres-tde-ext#helper-functions

documentation/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+* User data in tables, including TOAST tables, that are created using the extension. Metadata of those tables is not encrypted. 
+* Write-Ahead Log (WAL) data for tables created using the extension 
+* Temporary tables created during the database operation for data tables created using the extension
+
+## What's not encrypted
+
+In the MVP version of `pg_tde`, the following remains unencrypted:
+
+* Indexes
+* Logical replication
```
dAdAbird, line 22: Logical replication just doesn't work.

Logical replication uses WAL. As WAL is encrypted, logical replication routines are unable to de/en_crypt it properly to communicate data. This is something to be done after the MVP.

documentation/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+* `NULL` bitmaps of tuples
+* Keys in the keyring file
+
+Their encryption is planned for the next releases of `pg_tde`.
```
dAdAbird, line 26: We're planning to fix Logical replication and encrypt `NULL` bitmaps.

But I'm not sure we gonna do something with keys in the keyring file. It doesn't mean much sense - to encrypt it with what? With another key in some other filesystem file? Or using Vault? - then it makes sense to use Vault for keyring. Or am I missing something, @dutow?

Regarding indexes, I'm not sure how feasible to encrypt them with the current design, without changes in the PG core. I mean, we probably will encrypt indexes at some point. But it's not in our plans for now.
@codeforall @EngineeredVirus ^^^

documentation/docs/install.md
```diff
@@ -0,0 +1,69 @@
...
+        ```
+        git clone git://github.com/Percona-Lab/postgres-tde-ext
+        ```
+
+    5. Compile and install the extension
+
+        ```
+        cd postgres-tde-ext
+        make USE_PGXS=1
+        sudo make USE_PGXS=1 install
```
dAdAbird, line 28: After merging this PR https://github.com/Percona-Lab/postgres-tde-ext/pull/80
it is going to be: 
```suggestion
        cd postgres-tde-ext
        ./configure
        make USE_PGXS=1
        sudo make USE_PGXS=1 install
```

documentation/docs/index.md
dutow: We can (and probably will) add encrypted keyring file in the future, for example that's how keyring_aws works for mysql (it's basically a keyring_file encrypted with a key stored on amazon KMS)

documentation/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+
+<i warning>:material-alert: Warning:</i> Note that introducing encryption/decryption affects performance. Our benchmark tests show less than 10% performance overhead.
+
+[Get started](install.md){.md-button}
+
+## Supported PostgreSQL versions
+
+`pg_tde` is currently supported for Percona Distribution for PostgreSQL 16 and upstream PostgreSQL 16.
```
dutow, line 34: Since 16.1 is released, we should mention that it's currently based on 16.0.

documentation/docs/index.md
```diff
@@ -0,0 +1,40 @@
...
+* Temporary tables created during the database operation for data tables created using the extension
+
+## What's not encrypted
+
+In the MVP version of `pg_tde`, the following remains unencrypted:
+
+* Indexes
+* Logical replication
+* `NULL` bitmaps of tuples
+* Keys in the keyring file
```
dutow, line 24: After we merge the vault PR, keyring_file will be optional, and keys in vault will be secure.

documentation/docs/install.md
```diff
@@ -0,0 +1,70 @@
+# Installation
+
+Install `pg_tde` using one of available installation methods:
+
+=== "Build from source"
+
+    1. To build `pg_tde` from source code, you require the following on Ubuntu/Debian:
+
+        ```sh
+        sudo apt install make gcc libjson-c-dev postgresql-server-dev-16
```
dutow, line 10: After vault PR, we'll also need libcurl4-openssl-dev

documentation/docs/setup.md
```diff
@@ -0,0 +1,70 @@
...
+       sudo systemctl restart postgresql-16
+       ```
+
+3. Create the extension using the [CREATE EXTENSION](https://www.postgresql.org/docs/current/sql-createextension.html) command. You must have the privileges of a superuser or a database owner to use this command. Connect to `psql` as a superuser for a database and run the following command:
+
+    ```sql
+    CREATE EXTENSION pg_tde;
+    ```
+    
+    By default, the `pg_tde` extension is created for the `postgres` database or the database which your user owns. To encrypt the data in other databases, you must explicitly run the `CREATE EXTENSION` command for them.
```
dutow, line 31: "By default, the `pg_tde` extension is created for the `postgres` database or the database which your user owns." - this is not true, the extension will be created for the database the user executes the `CREATE EXTENSION pg_tde;` statement.

If you execute the `psql` command without a database name, it will try to connect to a database named the same as the current user - but we don't even mention the `psql` command here.

I think it would be better to just mention that the extension will only be created for the currently used database.

documentation/docs/setup.md
```diff
@@ -0,0 +1,70 @@
...
+## Keyring configuration
+
+```json
+{
+        'provider': 'file',
+        'datafile': '/tmp/pgkeyring',
+}
+```
+
+Currently the keyring configuration only supports the file provider, with a single datafile parameter.
```
dutow, line 66: Soon we'll also support vault, with the following extra configuration:

```json
{
        'provider': 'vault-v2',
        'token': 'ROOT_TOKEN',
        'url': 'http(s)://127.0.0.1:8200',
        'mountPath': 'secret',
        'caPath': '/optional/path/to/a/CaFile'
}
```

Here `token` is the secret access token to vault, which needs read/write access to the `mountPath` part of the vault server running at the `url`.

The vault keyring only supports version 2 of the KV secrets engine, which is the default for a long time now for vault.

(I'm also planning to add a mechanism for keeping the `token` more hidden, I'll add more details about that later)

documentation/docs/tde.md
```diff
@@ -0,0 +1,30 @@
+# What is Transparent Data Encryption (TDE)
+
+Transparent Data Encryption offers encryption at the file level and solves the problem of protecting data at rest. The encryption is completely transparent for users allowing them to access and manipulate the data and not to worry about the encryption process.
+
+## How does it work?
+
+In the MVP version of `pg_tde`, the keyring configuration file is used to store encryption keys.
```
dutow, line 7: This is not true, we are not storing keys in the configuration files. We are storing them in a datafile, if keyring_file is used.

documentation/docs/test.md
```diff
@@ -0,0 +1,13 @@
+# Test Transparent Data Encryption
+
+To check if the data is encrypted, do the following:
+
+1. Create a table in the database for which you have [enabled `pg_tde`](setup.md)
+2. Insert some test data
```
dutow, line 6: Inserting data is not required for this, that function just returns if the table uses pg_tde or not.

documentation/docs/test.md
```diff
@@ -0,0 +1,13 @@
+# Test Transparent Data Encryption
+
+To check if the data is encrypted, do the following:
```
dutow, line 3: This checks if a table is encrypted to not, it doesn't actually verify the data on disk - if somebody wants to do that, that can only be done manually.

documentation/docs/uninstall.md
```diff
@@ -0,0 +1,31 @@
+# Uninstall `pg_tde`
+
+If you no longer wish to use TDE in your deployment, you can remove the `pg_tde` extension. To do that, your user must have the privileges of the superuser or a database owner.
+
+Here's how to do it:
+
+1. Drop the extension using the `DROP EXTENSION` with `CASCADE` command
+
+   <i warning>:material-alert: Warning:</i> The use of the CASCADE parameter deletes all tables that were created in the database with `pg_tde` enabled.
```
dutow, line 9: And it also deletes anything else that depends on the encrypted tables. (For example, if a non encrypted table has a foreign key to an encrypted table, that non encrpyted table will be deleted too).

documentation/docs/uninstall.md
```diff
@@ -0,0 +1,31 @@
...
+
+       ```sh
+       sudo systemctl restart postgresql.service
+       ```
+    
+    * On RHEL and derivatives
+
+       ```sh
+       sudo systemctl restart postgresql-16
+       ```
```
dutow, line 31: At this point, the keyring configuration can be deleted too, and if keyring file is used, the datafile too.

documentation/docs/install.md
nastena1606: Added sine we will merge the vault PR for MVP, right?

documentation/docs/tde.md
nastena1606: Reworked

documentation/docs/install.md
dutow: yes

ImTheKai (approved):
LGTM

documentation/docs/contribute.md
```diff
@@ -0,0 +1,165 @@
+# Contributing guide
```
dAdAbird, line 1: Should we add a link to the Make Build developers docs: https://github.com/Percona-Lab/postgres-tde-ext/wiki/Make-builds-for-developers somewhere? Or just mention the WIKI in general as it expected to get more pages in future.

dAdAbird: Plus we'd probably want to mention that the project should follow Postgres' coding conventions (https://www.postgresql.org/docs/current/source.html). Although we don't follow them on our own. But have such intention)

documentation/docs/setup.md
nastena1606: Updated

documentation/docs/test.md
nastena1606: Removed

documentation/docs/contribute.md
dAdAbird: We do have tests. To run them:
```
./configure
make USE_PGXS=1 installcheck
```

dAdAbird (approved):

Issue #26, opened 2023-09-21
Add configure for make builds


dAdAbird:
For example, the include path of lib json-c on Mac is `/usr/local/include/json-c` as on Linux is `/usr/include/json-c`. Now it is hardcoded in Makefile. Snd Mac builds need changing Makefile on passing additional env during the compilation

dAdAbird:
https://github.com/Percona-Lab/postgres-tde-ext/commit/5279d3943bcf7be4ffce0c7fc4d996288743f39c

PR #80, opened 2023-12-01
Add configure to make builds

dAdAbird:
Trying to locate json-c library

`Makefile` is gone. It should be generated by configure based on `Makefile.in`

To `configre` in turn generated by `autoconf` based on `configure.ac` See https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/index.html

For https://github.com/Percona-Lab/postgres-tde-ext/issues/26


README.md
```diff
@@ -79,6 +79,7 @@ Compile and install the extension
 
 ```
 cd postgres-tde-ext
+./configure
```
dutow, line 82: We shouldn't commit `configure` directly to the repository, and document it as the first called script when using `autoconf`.

Similarly, `config.sub` and `config.guess` shouldn't be part of the commit, those are generated by `automake`.

Instead the build instructions should be something like:

```
cd postgres-tde-ext
autoconf
automake --add-missing
./configure
...
```

dAdAbird: I disagree. `configure` is a bash script that can be run anywhere. But `autoconf` and `automake` are tools that have to be installed and I don't see much reasons to impose these requirements and add extra steps to install procedure. 

```
./configure
make 
make install
```
Is pretty much standard "make" installation steps. Besides Postgres uses the same approach: https://www.postgresql.org/docs/current/install-make.html#INSTALL-SHORT-MAKE

I'd leave `autoconf` and `automake` only for developers when changing `configure.ac` (adding new required libs etc)

dutow: You are right about postgres, and looks like others follow this approach too - I never noticed this use of autoconf before.

But I still have two concerns with this:

1. Please document this somewhere visible, before one of us accidentally edits the configure script directly instead of the AC file
2. `config.sub` and `config.guess` are scripts that are updated regularly, and normally, automake generates these files in line with the installed linux/osx/... distribution. If we add them to our repository instead, we'll have to remember to update them regularly. Again, please document it somewhere - maybe we need a things to do before each "release" list or something, I'm not sure what to call it, as we do not yet have real releases.

capri1989: Please link from the Readme to the Wiki, so people don't miss it. Maybe @nastena1606 wants to update her bigger rework PR for the documentation

dutow (approved):

README.md
dAdAbird: https://github.com/Percona-Lab/postgres-tde-ext/wiki/Make-Builds-for-Developers. And added link to Readme

PR #84, opened 2023-12-07
WIP patch for sharing code - Not to be committed

EngineeredVirus:


src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+
+	if (lseek(keydata_file, write_pos, SEEK_SET) == -1)
+	{
+		ereport(FATAL,
+				(errcode_for_file_access(),
+					errmsg("Could not seek in tde key data file \"%s\": %m",
+						db_keydata_path)));
+	}
+
+	key_size = SizeOfRelKeysData(enc_rel_key_data->internal_keys_len);
```
dutow, line 484: The function description says that it only writes one key, but here it actually writes all of the internal keys for the table.

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+	int32 file_version;
+	char master_key_name[MASTER_KEY_NAME_LEN];
+} TDEFileHeader;
+
+typedef struct TDEMapEntry
+{
+	RelFileNumber relNumber;
+	int32 flags;
+	int32 key_index;
+} TDEMapEntry;
```
dutow, line 269: The comments say that flags/other data is in this map file, and the other file only contains the keys. However, currently that's not true, the other file also contains the start and end offsets together with the key.

With this design, it would make more sense to store the offsets in the map entry, and keep only the key itself in InternalKey, especially since the offsets can change during an internal key rotation, and we'll want to log those changes.

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+RelKeysData *
+pg_tde_get_key_from_file(const RelFileLocator *rlocator, const char *master_key_name)
+{
+	int32 key_index = 0;
+	const keyInfo *master_key_info;
+	RelKeysData *rel_key_data;
+	RelKeysData *enc_rel_key_data;
+
+	/* Get/generate a master, create the key for relation and get the encrypted key with bytes to write */
+	master_key_info = getMasterKey(master_key_name, true, true);
```
dutow, line 606: I think we discussed yesterday that we shouldn't create a master key here if one doesn't already exists, as that's an error?

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+
+	/* Get the file paths */
+	pg_tde_set_db_file_paths(rlocator);
+
+	/* Create the map entry and then add the encrypted key to the data file */
+	key_index = pg_tde_read_map_entry(rlocator, db_map_path, master_key_name);
+
+	/* Add the encrypted key to the data file. */
```
dutow, line 614: Incorrect comment? As we are reading here

src/access/pg_tde_tdemap.c
EngineeredVirus: Good point. I think we should remove the following members from the InternalKey structure as these are not being used anywhere in the code and I'm not sure if we'd ever use these.:

`
    Size    start_loc; 
    Size    end_loc;
	void*   ctx; // TODO: shouldn't be here / written to the disk
`

src/access/pg_tde_tdemap.c
EngineeredVirus: That's a valid point. I'll update the function to ensure that only one key is written.

On a side note, we need to have only one key in the RelKeysData structure. But that requires a longer discussion.

src/access/pg_tde_tdemap.c
EngineeredVirus: Good catch. The first argument should be false. And yes, based on yesterday's discussion, I've added a wrapper function.

src/access/pg_tde_tdemap.c
EngineeredVirus: Copy/paste :)

I've updated the comment.

src/access/pg_tde_tdemap.c
```diff
@@ -395,7 +664,9 @@ pg_tde_xlog_create_fork(XLogReaderState *record)
 {
 	char			*rec = XLogRecGetData(record);
 	RelFileLocator	rlocator;
-	InternalKey 	int_key = {0};
+	InternalKey 	int_key;
```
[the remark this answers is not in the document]
EngineeredVirus: Nothing wrong. Just an accidental change sneaked in while I was checking something.

src/encryption/enc_aes.c
```diff
@@ -43,6 +48,17 @@
...
+/* 
+ * Let's have a single definition of the IV across the code. SET_IV should
+ * be at the end of a variable declaration block. If not, a compilation
+ * warning will be generated informing of nonconformance ISO C90 due to
+ * mixing of declarations and code.
+ */
+typedef unsigned char TDE_iv[INTERNAL_KEY_LEN];
+#define SET_IV(iv)                                      \
+            TDE_iv iv;                                  \
+            memset((void *)iv, 0x00, sizeof(TDE_iv));
```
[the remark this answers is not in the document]
EngineeredVirus: For the time being, I think this is a fair change. Rather than having multiple copies of the IV in the code, this provides a cleaner view. We can always update as we change it later.

src/include/encryption/enc_aes.h
```diff
@@ -23,4 +25,7 @@ extern void Aes128EncryptedZeroBlocks(void* ctxPtr, const unsigned char* key, ui
 extern void AesEncrypt(const unsigned char* key, const unsigned char* iv, const unsigned char* in, int in_len, unsigned char* out, int* out_len);
 extern void AesDecrypt(const unsigned char* key, const unsigned char* iv, const unsigned char* in, int in_len, unsigned char* out, int* out_len);
 
+extern void AesEncryptKey(const keyInfo *master_key_info, RelKeysData *rel_key_data, RelKeysData *enc_rel_key_data, size_t *enc_key_bytes);
```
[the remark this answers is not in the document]
EngineeredVirus: These can always be updated later when we implement the IV change you mentioned above.

src/keyring/keyring_api.c
```diff
@@ -123,3 +122,22 @@ const keyInfo* keyringGenerateKey(const char* internalName, unsigned keyLen)
 	return keyringStoreKey(keyringConstructKeyName(internalName, i), kd);
 }
 
+const keyInfo* getMasterKey(const char* internalName, int doGenerateKey, int doRaiseError)
```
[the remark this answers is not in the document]
EngineeredVirus: I think it's better to keep getting the latest key and generation separate. The wrapper implies the code.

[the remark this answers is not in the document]
dutow: I agree, I meant that we could change the name of the function to follow the same style as the others.

src/access/pg_tde_tdemap.c
dutow: `start_loc` and `end_loc` will be used for internal key rotation. `ctx` is used, but shouldn't be written to disk, it is a runtime only variable.

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+	{
+		ereport(FATAL,
+				(errcode_for_file_access(),
+					errmsg("Could not seek in tde key data file \"%s\": %m",
+						db_keydata_path)));
+	}
+
+	key_size = SizeOfRelKeysData(enc_rel_key_data->internal_keys_len);
+
+	if (FileWrite(keydata_file, enc_rel_key_data, key_size, 0, WAIT_EVENT_DATA_FILE_WRITE) != key_size)
```
dAdAbird, line 486:
```suggestion
	if (FileWrite(keydata_file, enc_rel_key_data->internal_key, key_size, 0, WAIT_EVENT_DATA_FILE_WRITE) != key_size)
```
?

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+	File keydata_file = -1;
+	off_t write_pos;
+	bool is_new_file;
+	size_t key_size;
+
+	/* Open and vaidate file for basic correctness. */
+	keydata_file = pg_tde_open_file(db_keydata_path, master_key_name, O_RDWR | O_CREAT, &is_new_file);
+
+	/* Calculate the writing position in the file. */
+	write_pos = TDE_FILE_HEADER_SIZE + (key_index * INTERNAL_KEY_LEN);
```
dAdAbird, line 474: Shouldn't be here 
```suggestion
	write_pos = TDE_FILE_HEADER_SIZE + (key_index * SizeOfRelKeysData(1));
```
As we write whole `InternalKey` (with start_loc, end_loc etc) not just a key

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+	key_data_size = key_size - SizeOfRelKeysDataHeader;
+
+	/* Allocate and fill in the structure */
+	enc_rel_key_data = (RelKeysData *) palloc(key_size);
+
+	strcpy(enc_rel_key_data->master_key_name, master_key_name);
+	enc_rel_key_data->internal_keys_len = key_count;
+
+	/* Calculate the reading position in the file. */
+	read_pos = TDE_FILE_HEADER_SIZE + (key_index * INTERNAL_KEY_LEN);
```
dAdAbird, line 529:
```suggestion
	read_pos = TDE_FILE_HEADER_SIZE + (key_index * SizeOfRelKeysData(key_count));
```

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+
+	/*
+	 * Allocate in TopMemoryContext and don't pfree sice we add it to
+	 * the cache as well
+	 */
+	key_size = SizeOfRelKeysData(key_count);
+	key_data_size = key_size - SizeOfRelKeysDataHeader;
+
+	/* Allocate and fill in the structure */
+	enc_rel_key_data = (RelKeysData *) palloc(key_size);
```
dAdAbird, line 523: Comment above says `Allocate in TopMemoryContext` so here should be?
```suggestion
	enc_rel_key_data = (RelKeysData *) MemoryContextAlloc(TopMemoryContext, key_size);
```

src/access/pg_tde_tdemap.c
```diff
@@ -330,23 +177,445 @@ tde_sprint_masterkey(const keyData *k)
...
+	/* Get the file paths */
+	pg_tde_set_db_file_paths(rlocator);
+
+	/* Create the map entry and then add the encrypted key to the data file */
+	key_index = pg_tde_read_map_entry(rlocator, db_map_path, master_key_name);
+
+	/* Add the encrypted key to the data file. */
+	enc_rel_key_data = pg_tde_read_keydata(db_keydata_path, key_index, master_key_info->name.name);
+	rel_key_data = tde_decrypt_rel_key(master_key_info, enc_rel_key_data);
+
```
dAdAbird, line 617: I think we can `pfree(enc_rel_key_data)` now.

src/access/pg_tde_tdemap.c
dAdAbird: > On a side note, we need to have only one key in the RelKeysData structure. But that requires a longer discussion.

Maybe it worth adding some comment that although `RelKeysData` suppose multiple internal keys, currently we always write and read only one key. I mean the calculation of read/write pos in pg_tde_read/write_keydata() always suppose one key. Add looking for a free spot for only one internalKey in pg_tde_write_map_entry()

dAdAbird (changes requested):

PR #91, opened 2023-12-18
Implementing remote options for keyring configuration

dutow:
This commit adds the possibility to extract parameters from the main keyring configuration file into a separate file, or remote HTTP objects. Basic tests using perl are also added.

Two simple configuration files are 'keyring-w-file.json' and 'keyring-w-http.json'. Additionally, the HTTP version also support a 'ca' parameter for SSL, similar to the main keyring vault configuration.

With this commit, all string parameters support these options, except the main 'provider' option.


dAdAbird (commented):
Should we expand [the configuration section of Readme](https://github.com/Percona-Lab/postgres-tde-ext#keyring-configuration)?

dAdAbird (commented):
Just a one minor comment, otherwise LGTM

src/catalog/tde_keyring.c
```diff
@@ -224,31 +226,167 @@ load_keyring_provider_options(ProviderType provider_type, Datum keyring_options)
 	return NULL;
 }
 
+static const char *
```
codeforall, line 229: We should move these extract_json_cstr() and extract_json_option_value() utility functions to a separate file. These don't seem to belong to the keyring catalog. I think moving these to pg_tde_utils.c file could be a good option.

src/catalog/tde_keyring.c
```diff
@@ -224,31 +226,167 @@ load_keyring_provider_options(ProviderType provider_type, Datum keyring_options)
...
+		return extract_json_cstr(top_json, field_name);
+	}
+
+	if(strcmp(field_type_cstr, "object") != 0)
+	{
+		elog(ERROR, "Unsupported type for object %s: %s", field_name, field_type_cstr);
+		return NULL;
+	}
+
+	// Now it is definitely an object
```
codeforall, line 266: This and there are few other single line comments in the code

src/catalog/tde_keyring.c
```diff
@@ -224,31 +226,167 @@ load_keyring_provider_options(ProviderType provider_type, Datum keyring_options)
...
+
+			f = fopen(path_cstr, "r");
+
+			if(!f)
+			{
+				elog(ERROR, "The file referenced by %s doesn't exists, or is not readable to postgres: %s", field_name, path_cstr);
+				return NULL;
+			}
+
+			out = palloc(1024);
```
codeforall, line 339: Define instead of 1024? And some comments, what we expect to read from the file

dutow: Done.

src/catalog/tde_keyring.c
dutow: Done

src/catalog/tde_keyring.c
dutow: Done

codeforall (approved):

PR #92, opened 2023-12-18
Simple test job to test that docker image builds and works

dutow:
With this additional github action, all pull request try to build and use the docker image as part of the CI tests, to ensure that we do not break it without noticing.


dAdAbird (approved):
Just thoughts:
Can we set the order of jobs? Maybe it make sense to run this job after the publication of images on docker hub and use that image here instead of building. So we kill two birds with one stone also checking the docker-hub image.
On the other hand it makes sense to keep the testing scope narrow. 
Either way works for me, LGTM

dutow:
Jobs can have dependencies, but that job only runs on the main branch after we merged a PR, and this job runs on a PR branch after we create/update the PR. It's also possible that a build will fail, in which case we wouldn't push anything to dockerhub.

EngineeredVirus (approved):
LGTM

PR #93, opened 2023-12-18
Added a note about table access methods

### LINKED ISSUE #87: It is not clear that the user has to explicitly set the access method
I followed the instructions well enough to have the database up to run some tests, after realizing I had no performance drop I started reviewing the docs, when I got to the end of ["Test Transparent Data Encryption"](https://percona-lab.github.io/postgres-tde-ext/main/test.html) I got stuck for a while, from the docs it seems that I did everything right.

Maybe I missed it in some other place but I saw no mention that I had to set the access method for my test table explicitly, I assumed that the extension itself would set `default_table_access_method` and did not check. 

I believe that other users not so familiar with access methods would struggle.

nastena1606:
Changed GH action to trigger only t=on changes to docs folder

modified:   .github/workflows/doc-build.yaml
	modified:   documentation/docs/test.md

### DIFF
--- a/.github/workflows/doc-build.yaml
+++ b/.github/workflows/doc-build.yaml
@@ -3,7 +3,8 @@ on:
   push:
     branches: 
       - main
-      - Docs
+    paths:
+      - documentation
      
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


dutow (approved):

ImTheKai (approved):
LGTM

Issue #90, opened 2023-12-16
using vault in tde_conf.json sends wrong http requests to vault


Torkolis:
So I have managed to set up pg_tde using vault. Basically the 'mountPath' entry needs to match a mount of the kv secret engine (explanation in the docs would be cool, took me a while to figure it out). So when I created the first table it wrote a secret in that entry called "master-key-1" containing some data:
![image](https://github.com/Percona-Lab/postgres-tde-ext/assets/98939160/e1afcb5f-881b-4b69-8618-457bbc41578e)
But now I have a problem that it tries to access master-key-2 in vault which does not exist and returns 404. The usual database encryption actually works as expected, but I just don't understand why it makes GET requests to master-key-2... I think it is probably a bug. I did some tests and here are the nginx logs (my vault reachable behind an nginx):

```
create first table
<IP> - - [16/Dec/2023:10:21:10 +0000] "GET /v1/secret/data/master-key-1 HTTP/1.1" 404 14 "-" "-" 180 0.002 [vault-vault-8200] [] <another_IP>:8200 14 0.002 404 fc88f07d2dbbe795aa8b830bcb96f7b9
<IP> - - [16/Dec/2023:10:21:10 +0000] "GET /v1/secret/data/master-key-1 HTTP/1.1" 404 14 "-" "-" 180 0.001 [vault-vault-8200] [] <another_IP>:8200 14 0.001 404 1b02460c5b4561f8d52d892193ec4b4f
<IP> - - [16/Dec/2023:10:21:10 +0000] "POST /v1/secret/data/master-key-1 HTTP/1.1" 200 279 "-" "-" 250 0.001 [vault-vault-8200] [] <another_IP>:8200 279 0.001 200 067a6fe51844a7bba789a22326af51b9
<IP> - - [16/Dec/2023:10:21:10 +0000] "GET /v1/secret/data/master-key-2 HTTP/1.1" 404 14 "-" "-" 180 0.000 [vault-vault-8200] [] <another_IP>:8200 14 0.000 404 64123d6ceda08c8d5efd126ca52cc956
insert into first table
<IP> - - [16/Dec/2023:10:22:02 +0000] "GET /v1/secret/data/master-key-2 HTTP/1.1" 404 14 "-" "-" 180 0.001 [vault-vault-8200] [] <another_IP>:8200 14 0.001 404 ee2228486eb7d2a79d01e451c80832d1
create second table
[...]

PR #95, opened 2023-12-19
Store proper name with version in fork file

### LINKED ISSUE #94: Fork files store just generic key name, not specific version
Fork files should store a specific key version, for example "master-key-13", but currently only store the generic name, "master-key"

dutow:
Currently the fork file stores just the string "master-key", which is incorrect, as it does not contain the version information. To counteract this, at read time it loads the latest master key, but that can be different than the version which was used to encrypt it.

These getLatestKey calls also lead to postgres making many unneccessary calls to vault, making this bug partially responsible for #90.

The fix is simple, from now on we correctly write the entire string including the version number, and read the specified version at read time.

Fixes #94.


dAdAbird (approved):

EngineeredVirus (commented):
Why the use of strcpy at one place and strncpy at the other? Also, I think we should replace 256 with MASTER_KEY_NAME_LEN or another define.

EngineeredVirus (approved):
LGTM

PR #97, opened 2024-01-10
Encrypt relation's Internal Key in XLog

dAdAbird:
XLog record with its Internal Key for replication purposes. This commit makes this Key encrypted with the master key. Therefore we have to add also a master key name to XLog. That also means that standbys should have access to the very same master key as the primary.

Fix https://github.com/Percona-Lab/postgres-tde-ext/issues/56

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -34,8 +34,9 @@ static const char *MasterKeyName = "master-key";
 
 static inline char* pg_tde_get_key_file_path(const RelFileLocator *newrlocator);
 static void put_keys_into_map(Oid rel_id, RelKeysData *keys);
-static void pg_tde_write_key_fork(const RelFileLocator *rlocator, InternalKey *key, const char *MasterKeyName);
+static void pg_tde_add_rel_keys(const RelFileLocator *rlocator, InternalKey *key, InternalKey *keyEnc, const char *masterkeyname);
 static void pg_tde_xlog_create_fork(XLogReaderState *record);
+static void pg_tde_decrypt_internal_key(const InternalKey *in, InternalKey *out, const char *masterkeyname);
 
 void
 pg_tde_delete_key_fork(Relation rel)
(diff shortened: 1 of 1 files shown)


src/access/pg_tde_tdemap.c
```diff
@@ -68,45 +76,50 @@ pg_tde_create_key_fork(const RelFileLocator *newrlocator, Relation rel)
...
 	}
 
+	master_key_info = keyringGetLatestKey(MasterKeyName);
+	if(master_key_info == NULL)
+	{
+		master_key_info = keyringGenerateKey(MasterKeyName, INTERNAL_KEY_LEN);
+	}
+	if(master_key_info == NULL)
+	{
+        ereport(ERROR,
```
dutow, line 86: Incorrect indentation - it's already there in the original, but now that we move it, it's a good time to fix it

src/access/pg_tde_tdemap.c
```diff
@@ -394,6 +398,9 @@ pg_tde_xlog_create_fork(XLogReaderState *record)
 	char			*rec = XLogRecGetData(record);
 	RelFileLocator	rlocator;
 	InternalKey 	int_key = {0};
+	InternalKey 	int_key_enc = {0};
+	uint8 			mkey_len;
```
dutow, line 402: seems like another indentation issue?

src/access/pg_tde_tdemap.c
```diff
@@ -347,6 +330,27 @@ pg_tde_get_key_file_path(const RelFileLocator *newrlocator)
     return key_file_path;
 }
 
+void 
+pg_tde_decrypt_internal_key(const InternalKey *in, InternalKey *out, const char *MasterKeyName)
```
dutow, line 334: `MasterKeyName` also seems inconsistent with the other variable/argument names

dutow (commented):
Looks good, I only added a few formatting related comments

dAdAbird:
I reckoned it make sense to run `pgindent` on the file

dutow:
@dAdAbird pgIndent is a good idea, but could we do it in a separate PR instead? Maybe even on the entire code?

With this now, if you squash and merge, we'll end up with lots of formatting changes and some logic changes in a single commit.

dutow (approved):

EngineeredVirus (approved):
LGTM

PR #99, opened 2024-01-15
Fix json syntax

artemgavrilov:
JSON format requires double quotes and last element should not end with a comma
https://www.json.org/json-en.html

>A string is a sequence of zero or more Unicode characters, wrapped in double quotes, using backslash escapes.


nastena1606 (approved):
LGTM, thanks!

dAdAbird (approved):

Issue #25, opened 2023-09-21
Broken meson builds on Mac


dAdAbird:
Linking during meson builds is failing with:
```
[5/6] Linking target contrib/postgres-tde-ext/enc_test
FAILED: contrib/postgres-tde-ext/enc_test
cc  -o contrib/postgres-tde-ext/enc_test contrib/postgres-tde-ext/enc_test.p/src_encryption_enc_aes.c.o contrib/postgres-tde-ext/enc_test.p/src_encryption_test.c.o -L/usr/local/Cellar/openssl@3/3.1.2/lib -I/usr/local/Cellar/openssl@3/3.1.2/include -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -Wl,-undefined,error -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.0.sdk -Wl,-rpath,/usr/local/Cellar/lz4/1.9.4/lib -Wl,-rpath,/usr/local/Cellar/openssl@3/3.1.2/lib -Wl,-rpath,/usr/local/Cellar/zstd/1.5.5/lib -Wl,-rpath,/usr/local/Cellar/json-c/0.17/lib -Wl,-undefined,error -fvisibility=hidden -bundle_loader /Users/andrew/projects/PG/postgres/build/src/backend/postgres -lldap -lxml2 /usr/local/Cellar/lz4/1.9.4/lib/liblz4.dylib -lpam /usr/local/Cellar/openssl@3/3.1.2/lib/libssl.dylib /usr/local/Cellar/openssl@3/3.1.2/lib/libcrypto.dylib -lz /usr/local/Cellar/zstd/1.5.5/lib/libzstd.dylib -lm /usr/local/Cellar/json-c/0.17/lib/libjson-c.dylib
ld: warning: -undefined error is deprecated
ld: warning: -undefined error is deprecated
ld: -bundle_loader can only be used with -bundle
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[6/6] Linking target contrib/postgres-tde-ext/pg_tde.dylib
ld: warning: -undefined error is deprecated
ninja: build stopped: subcommand failed.
```

It has to do with unit tests in `meson.build`. As it compiles fine if to remove [this](https://github.com/Percona-Lab/postgres-tde-ext/blob/a3d687271faab7da97caee58304e6e104aa354f9/meson.build#L44-L48)

PR #100, opened 2024-01-17
Add MacOS specific linker flags to meson tests build step for MacOS

artemgavrilov:
Fixes https://github.com/Percona-Lab/postgres-tde-ext/issues/25

Unlike linux on MacOS [shared library and loadable module(bundle)](https://www.oreilly.com/library/view/mac-os-x/0596003560/ch05s03.html) are different things. So linker was miss `-bundle` flag for testing executables build.

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/LoadingCode/Concepts/AboutLoadableBundles.html

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -43,11 +43,19 @@ pg_tde = shared_module('pg_tde',
 )
 contrib_targets += pg_tde
 
+ldflags = []
+if host_machine.system() == 'darwin'
+  # On MacOS Shared Libraries and Loadable Modules are different things,
+  # so we need to pass an extra flag to the linker.
+  ldflags += '-bundle'
+endif
+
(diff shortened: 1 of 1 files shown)


dAdAbird (approved):
LGTM

dutow:
@artemgavrilov could you also try to add a macos based github action? It can be a separate PR, but it would help with keeping the build working in the future.

artemgavrilov:
> @artemgavrilov could you also try to add a macos based github action? It can be a separate PR, but it would help with keeping the build working in the future.

Sure, will do it in separate PR

codeforall (approved):
Looks good

Issue #101, opened 2024-01-18
Fix TOAST Initialization vector


dAdAbird:
Currently we encrypt TOASTed data always with the offset 0. That is not secure. The offset should be unique.

PR #102, opened 2024-01-18
Fix TOAST Initialization vector

dAdAbird:
Currently, we encrypt TOASTed data always with the offset 0. That is not secure. The offset should be unique.

This commit replaces the 0 "offset" with TOAST's `va_valueid` (Unique ID of value within the TOAST table) during encryption. This `va_valueid` is available during the TOAST fetch which is crucial for the decryption.

During the TOAST externalisation we insert a new tuple which shouldn't be encrypted as the backend will give this tuple to us during the TOAST fetch, hence fetched with non-TDE functions, besides TOAST data already encrypted. For that (insert non-encrypted tuple) I had to modify some TDE AM functions.

`pg_tde_toast_save_datum()` was copied from the PG code and modified. Along with `toastrel_valueid_exists()` and `toastid_valueid_exists()`.

Fix https://github.com/Percona-Lab/postgres-tde-ext/issues/101

### DIFF
--- a/src/access/pg_tdetoast.c
+++ b/src/access/pg_tdetoast.c
@@ -32,13 +32,23 @@
 #include "access/genam.h"
 #include "access/toast_helper.h"
 #include "access/toast_internals.h"
+#include "miscadmin.h"
 #include "utils/fmgroids.h"
+#include "utils/snapmgr.h"
 #include "encryption/enc_tuple.h"
 
 #define TDE_TOAST_COMPRESS_HEADER_SIZE (VARHDRSZ_COMPRESSED - VARHDRSZ)
 
-static void
-pg_tde_toast_tuple_externalize(ToastTupleContext *ttc, int attribute, int options);
+static void pg_tde_toast_tuple_externalize(ToastTupleContext *ttc,
+									int attribute, int options);
+static Datum pg_tde_toast_save_datum(Relation rel, Datum value,
+								struct varlena *oldexternal,
+								int options);
+static void pg_tde_toast_encrypt(Pointer dval, Oid valueid, RelKeysData *keys);
+static bool toastrel_valueid_exists(Relation toastrel, Oid valueid);
+static bool toastid_valueid_exists(Oid toastrelid, Oid valueid);
+
+
 /* ----------
  * pg_tde_toast_delete -
  *
@@ -799,7 +809,7 @@ pg_tde_fetch_toast_slice(Relation toastrel, Oid valueid, int32 attrsize,
 			}
 		}
 		/* Decrypt the data chunk by chunk here */
-		PG_TDE_DECRYPT_DATA((curchunk * TOAST_MAX_CHUNK_SIZE - sliceoffset) + encrypt_offset,
+		PG_TDE_DECRYPT_DATA((curchunk * TOAST_MAX_CHUNK_SIZE - sliceoffset) + encrypt_offset + valueid,
 					chunkdata + chcpystrt,
 					(chcpyend - chcpystrt) + 1,
 					decrypted_data, keys);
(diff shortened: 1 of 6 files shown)


src/include/access/pg_tdeam.h
```diff
@@ -236,7 +236,7 @@ extern BulkInsertState GetBulkInsertState(void);
 extern void FreeBulkInsertState(BulkInsertState);
 extern void ReleaseBulkInsertStatePin(BulkInsertState bistate);
 
-extern void pg_tde_insert(Relation relation, HeapTuple tup, CommandId cid,
+extern void pg_tde_insert(bool encrypt, Relation relation, HeapTuple tup, CommandId cid,
```
codeforall, line 239: What if we use the flag 'int options' argument of the function to indicate if the tuple needs to be encrypted or not instead of adding a new argument, Currently there are only four possible flags that can be passed through 'options field', and we have a penalty of bits available to define a new flag?

src/access/pg_tdetoast.c
```diff
@@ -865,10 +862,381 @@ pg_tde_toast_tuple_externalize(ToastTupleContext *ttc, int attribute, int option
...
+	*value = pg_tde_toast_save_datum(ttc->ttc_rel, old_value, attr->tai_oldexternal,
+							  options);
+	if ((attr->tai_colflags & TOASTCOL_NEEDS_FREE) != 0)
+		pfree(DatumGetPointer(old_value));
+	attr->tai_colflags |= TOASTCOL_NEEDS_FREE;
+	ttc->ttc_flags |= (TOAST_NEEDS_CHANGE | TOAST_NEEDS_FREE);
+}
+
+/* ----------
+ * toast_save_datum -
```
codeforall, line 891: Function comments are still using the old function function. Let's reference the toast_save_datum() function here; we have copied this function from the PG code and made the following changes. I think this will be helpful when we merge the changes made to toast_save_datum() in the PG code.

src/include/access/pg_tdeam.h
dAdAbird: great idea, I'll take a look, thanks

src/access/pg_tdetoast.c
dAdAbird: https://github.com/Percona-Lab/postgres-tde-ext/pull/102/commits/5f98c7359f198d77537c5123ae8595f5c47465d3

src/include/access/pg_tdeam.h
dAdAbird: https://github.com/Percona-Lab/postgres-tde-ext/pull/102/commits/5f98c7359f198d77537c5123ae8595f5c47465d3

dutow (commented):
I have two questions: 

* can we also guarantee that we won't have any overlaps with `valueid`? (if not, we can fix that after the changes in my next PR - we can do that later, the question if we have to do that or not)
* wouldn't it be better to split it into two commits, one for copying the core code, one for the changes in them? That would make it cleaner what are our changes

src/access/pg_tdetoast.c
```diff
@@ -865,10 +867,390 @@ pg_tde_toast_tuple_externalize(ToastTupleContext *ttc, int attribute, int option
...
+	struct varlena *result;
+	struct varatt_external toast_pointer;
+	union
+	{
+		struct varlena hdr;
+		/* this is to make the union big enough for a chunk: */
+		char		data[TOAST_MAX_CHUNK_SIZE + VARHDRSZ];
+		/* ensure union is aligned well enough: */
+		int32		align_it;
+	}			chunk_data;
```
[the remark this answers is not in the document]
codeforall: Having an int (int 32 align_it) in the union ensures that the union variable will always start from the aligned address. Without this int (since all other members in the union and varlina structure are chars), the union variable can be placed at an unaligned starting address. align_it in the union makes sure the starting address is always aligned no matter the size of the union.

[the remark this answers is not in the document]
dAdAbird: I think this is legacy from times when varlena was 5bytes: https://github.com/postgres/postgres/blob/c67f6f2f573064c206044b44a73cdf0806dfbd4e/src/include/c.h#L411-L415. At least at times when this padding was introduced: https://github.com/postgres/postgres/commit/c67f6f2f573064c206044b44a73cdf0806dfbd4e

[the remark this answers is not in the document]
codeforall: I believe it's more to guard against stack variables of chunk_data to start at an aligned address, and the padding is still required.

codeforall (approved):
Looks perfect now

src/access/pg_tdetoast.c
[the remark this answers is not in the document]
dutow: For some reason I thought that `varlena` is a pointer there, but of course it's not, it's clear then.

dutow (approved):

Issue #82, opened 2023-12-04
Updating the toasted column in encrypted table causes the crash


codeforall:
Trying to update a toasted column causes the server crash. Only happens when there is another column proceeds the toasted column

```

DROP TABLE indtoasttest;
CREATE TABLE indtoasttest(descr text, cnt int DEFAULT 0, f1 text, f2 text) using pg_tde;
INSERT INTO indtoasttest(descr, f1, f2) VALUES('two-compressed', repeat('1234567890',1000), repeat('1234567890',1000));
INSERT INTO indtoasttest(descr, f1, f2) VALUES('two-toasted', repeat('1234567890',30000), repeat('1234567890',50000));
INSERT INTO indtoasttest(descr, f1, f2) VALUES('one-compressed,one-null', NULL, repeat('1234567890',1000));
INSERT INTO indtoasttest(descr, f1, f2) VALUES('one-toasted,one-null', NULL, repeat('1234567890',50000));
UPDATE indtoasttest SET cnt = cnt +1 RETURNING substring(indtoasttest::text, 1, 200);
UPDATE indtoasttest SET cnt = cnt +1, f1 = f1 RETURNING substring(indtoasttest::text, 1, 200);
UPDATE indtoasttest SET cnt = cnt +1, f1 = f1||'' RETURNING substring(indtoasttest::text, 1, 200);
UPDATE indtoasttest SET cnt = cnt +1, f1 = f1||'' RETURNING substring(indtoasttest::text, 1, 200);
UPDATE indtoasttest SET  f2 = '+'||f2||'-' ;


```

PR #103, opened 2024-01-19
Updating the toasted column in encrypted table causes the crash (#82)

codeforall:
pg_tde_toast_insert_or_update calls pg_tde_deform_tuple  on old-tuple from within and tries to dereference the toast columns. Passing the encrypted tuple as part of the old tuple renders the toast data pointer (the encrypted value of the actual pointer) invalid, and anything can happen if it gets dereferenced as it is.
The solution is to pass the decrypted old tuple pg_tde_toast_insert_or_update function from pg_tde_update.

### DIFF
--- a/src/access/pg_tdeam.c
+++ b/src/access/pg_tdeam.c
@@ -3634,7 +3634,7 @@ l2:
 		if (need_toast)
 		{
 			/* Note we always use WAL and FSM during updates */
-			heaptup = pg_tde_toast_insert_or_update(relation, newtup, &oldtup, 0);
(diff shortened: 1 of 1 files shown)


dAdAbird (commented):
Can you add test for this? Anyway there is a test scenario in https://github.com/Percona-Lab/postgres-tde-ext/issues/82

And also reference https://github.com/Percona-Lab/postgres-tde-ext/issues/82 issue in the commit/PR message

dAdAbird (approved):

EngineeredVirus (approved):
LGTM

PR #105, opened 2024-01-22
Fixing errors with OpenSSL 1.0/1.1

dutow:
Issue: there's a slight behavior difference between OpenSSL 1.1 and 3.0 in regards to padding, and how the EVP encryption/decryption functions exactly behave.

Our code also incorrectly used these functions until now, which by chance worked correctly with 3.0, but caused decryption issues with 1.0/1.1.

This commit:

* Fixes the issue
* Adds a few related assertions
* Adds an Ubuntu 20.04 (OpenSSL 1.1) github runner to verify that everything works correctly with 1.1.

### DIFF
--- a/src/encryption/enc_aes.c
+++ b/src/encryption/enc_aes.c
@@ -107,9 +107,6 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 	ctx = EVP_CIPHER_CTX_new();
 	EVP_CIPHER_CTX_init(ctx);
 
-	EVP_CIPHER_CTX_set_padding(ctx, 0);
-	Assert(in_len % cipher_block_size == 0);
-
 	if(EVP_CipherInit_ex(ctx, cipher, NULL, key, iv, enc) == 0)
 	{
 		#ifdef FRONTEND
@@ -120,6 +117,9 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 		#endif
 		goto cleanup;
 	}
+
+	EVP_CIPHER_CTX_set_padding(ctx, 0);
+	Assert(in_len % cipher_block_size == 0);
 
 
 	if(EVP_CipherUpdate(ctx, out, out_len, in, in_len) == 0)
--- a/src/encryption/enc_aes.c
+++ b/src/encryption/enc_aes.c
@@ -102,11 +102,13 @@ AesRun2(EVP_CIPHER_CTX** ctxPtr, int enc, const unsigned char* key, const unsign
 
 static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, const unsigned char* in, int in_len, unsigned char* out, int* out_len)
 {
+	int out_len2 = 0;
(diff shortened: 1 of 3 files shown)


src/encryption/enc_aes.c
```diff
@@ -119,6 +120,9 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 		goto cleanup;
 	}
 
+	EVP_CIPHER_CTX_set_padding(ctx, 0);
```
EngineeredVirus, line 123: We are already calling EVP_CIPHER_CTX_set_padding up above. Is the earlier call redundant now?

src/encryption/enc_aes.c
```diff
@@ -141,6 +145,9 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 		goto cleanup;
 	}
 
+	*out_len += out_len2;
+	Assert(in_len == *out_len);
```
EngineeredVirus, line 149: Do you think this is a valid assertion? Are there any scenarios where EVP_CipherFinal_ex returns 1, and yet we can end up with out_len not equal to in_len?

src/encryption/enc_aes.c
dutow: No, the above call was before we initialized the context, I deleted it.

src/encryption/enc_aes.c
```diff
@@ -102,12 +102,11 @@ AesRun2(EVP_CIPHER_CTX** ctxPtr, int enc, const unsigned char* key, const unsign
 
 static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, const unsigned char* in, int in_len, unsigned char* out, int* out_len)
 {
+	int out_len2 = 0;
```
EngineeredVirus, line 105: We need some comments about why out_len2 is required so that people who don't understand SSL can also understand the changes.

dutow: I already added a comment around the second assertion

src/encryption/enc_aes.c
```diff
@@ -119,6 +118,10 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 		goto cleanup;
 	}
 
+	EVP_CIPHER_CTX_set_padding(ctx, 0);
+	Assert(in_len % cipher_block_size == 0);
+
```
EngineeredVirus, line 123: Two empty new lines. When we run pgindent, it should take care of this, however, we should try to avoid whitespace issues per pg coding standards.

src/encryption/enc_aes.c
```diff
@@ -141,6 +144,11 @@ static void AesRun(int enc, const unsigned char* key, const unsigned char* iv, c
 		goto cleanup;
 	}
 
+	// We encrypt one block (16 bytes)
```
EngineeredVirus, line 147: Better to use multiline comments /* ... */ instead of single line ones.

EngineeredVirus (approved):
LGTM

codeforall (approved):
Looks good

PR #113, opened 2024-01-30
Key Map Implementation - Replacing TDE Forks

EngineeredVirus:
This PR has two commits with one commit including some refactoring, and the other commit includes implementation of the key map architecture. Unfortunately, since most of this code was written back in December, it is very hard to recreate a separate PR for the refactoring part. Therefore, for ease, I've created a separate commit for it.

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -137,9 +137,7 @@ pg_tde_create_key_map_entry(const RelFileLocator *newrlocator, Relation rel)
 	XLogInsert(RM_TDERMGR_ID, XLOG_TDE_RELATION_KEY);
 
 	/*
-	 * TODO: should DB crash after sending XLog, secondaries would create a fork
-	 * file but the relation won't be created either on primary or secondaries.
-	 * Hence, the *.tde file will remain as garbage on secondaries.
+	 * Add the encyrpted key to the key map data file structure.
 	 */
 	pg_tde_write_key_map_entry(newrlocator, enc_rel_key_data, master_key_info);
 }
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -81,7 +81,7 @@ static const char *MasterKeyName = "master-key";
 static void put_keys_into_map(Oid rel_id, RelKeysData *keys);
 static void pg_tde_xlog_create_relation(XLogReaderState *record);
 
-static RelKeysData* tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const keyInfo *master_key_info);
+static RelKeysData* tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const keyInfo *master_key_info, bool is_key_decrypted);
 static RelKeysData* tde_encrypt_rel_key(const keyInfo *master_key_info, RelKeysData *rel_key_data);
 static RelKeysData* tde_decrypt_rel_key(const keyInfo *master_key_info, RelKeysData *enc_rel_key_data);
 static bool pg_tde_perform_rotate_key(const char *new_master_key_name);
@@ -127,7 +127,7 @@ pg_tde_create_key_map_entry(const RelFileLocator *newrlocator, Relation rel)
 	}
 
 	/* Encrypt the key */
-	rel_key_data = tde_create_rel_key(newrlocator, &int_key, master_key_info);
+	rel_key_data = tde_create_rel_key(newrlocator, &int_key, master_key_info, true);
 	enc_rel_key_data = tde_encrypt_rel_key(master_key_info, rel_key_data);
 
 	/* XLOG internal keys */
@@ -219,18 +219,22 @@ tde_sprint_masterkey(const keyData *k)
  * created key.
  */
 RelKeysData *
-tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const keyInfo *master_key_info)
+tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const keyInfo *master_key_info, bool is_key_decrypted)
 {
 	RelKeysData 	*rel_key_data;
+	MemoryContext context = (is_key_decrypted ? TopMemoryContext : CurrentMemoryContext);
 
-	rel_key_data = (RelKeysData *) MemoryContextAlloc(TopMemoryContext, SizeOfRelKeysData(1));
+	Assert(context);
+
+	rel_key_data = (RelKeysData *) MemoryContextAlloc(context, SizeOfRelKeysData(1));
 
 	strncpy(rel_key_data->master_key_name, master_key_info->name.name, MASTER_KEY_NAME_LEN);
-	rel_key_data->internal_key[0] = *key;
+	memcpy(&rel_key_data->internal_key[0], key, sizeof(InternalKey));
 	rel_key_data->internal_keys_len = 1;
 
-	/* Add to the cache */
-	put_keys_into_map(rlocator->relNumber, rel_key_data);
+	/* Add to the decrypted key to cache */
+	if (is_key_decrypted)
+		put_keys_into_map(rlocator->relNumber, rel_key_data);
 
 	return rel_key_data;
 }
@@ -702,7 +706,7 @@ pg_tde_write_key_map_entry(const RelFileLocator *rlocator, RelKeysData *enc_rel_
 {
 	int32 key_index = 0;
 
-	/* Get the file paths */
+	/* Set the file paths */
 	pg_tde_set_db_file_paths(rlocator, NULL);
 
 	/* Create the map entry and then add the encrypted key to the data file */
@@ -921,7 +925,10 @@ pg_tde_xlog_create_relation(XLogReaderState *record)
 	master_key_info = getMasterKey(MasterKeyName, true, true);
 
 	/* Get the the keydata structure for the encrypted key */
-	enc_rel_key_data = tde_create_rel_key(&rlocator, &int_key, master_key_info);
+	enc_rel_key_data = tde_create_rel_key(&rlocator, &int_key, master_key_info, true);
+
+	/* Add the key to key cache */
+	tde_create_rel_key(&rlocator, &enc_rel_key_data->internal_key[0], master_key_info, false);
 
 	pg_tde_write_key_map_entry(&rlocator, enc_rel_key_data, master_key_info);
 }
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -24,6 +24,7 @@
 
 #include "access/pg_tde_tdemap.h"
 #include "encryption/enc_aes.h"
+#include "encryption/enc_tde.h"
 #include "keyring/keyring_api.h"
 
 #include <openssl/rand.h>
@@ -78,7 +79,7 @@ static char db_keydata_path[MAXPGPATH] = {0};
 static const char *MasterKeyName = "master-key";
 
 static void put_keys_into_map(Oid rel_id, RelKeysData *keys);
-static void pg_tde_xlog_create_fork(XLogReaderState *record);
+static void pg_tde_xlog_create_relation(XLogReaderState *record);
 
 static RelKeysData* tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const keyInfo *master_key_info);
 static RelKeysData* tde_encrypt_rel_key(const keyInfo *master_key_info, RelKeysData *rel_key_data);
@@ -87,7 +88,7 @@ static bool pg_tde_perform_rotate_key(const char *new_master_key_name);
 
 static void pg_tde_set_db_file_paths(const RelFileLocator *rlocator, char *str_append);
 static File pg_tde_open_file(char *tde_filename, const char *master_key_name, int fileFlags, bool *is_new_file, off_t *offset);
-static void pg_tde_write_key_map_entry(const RelFileLocator *rlocator, InternalKey *key, const char *master_key_name);
+static void pg_tde_write_key_map_entry(const RelFileLocator *rlocator, RelKeysData *enc_rel_key_data, const keyInfo *master_key_info);
 
 static int32 pg_tde_write_map_entry(const RelFileLocator *rlocator, char *db_map_path, const char *master_key_name);
 static int32 pg_tde_write_one_map_entry(File map_file, const RelFileLocator *rlocator, int flags, int32 key_index, TDEMapEntry *map_entry, off_t *offset);
@@ -108,6 +109,12 @@ void
 pg_tde_create_key_map_entry(const RelFileLocator *newrlocator, Relation rel)
 {
 	InternalKey int_key;
+	RelKeysData *rel_key_data;
+	RelKeysData *enc_rel_key_data;
+	const keyInfo *master_key_info;
+
+	/* Get/generate a master, create the key for relation and get the encrypted key with bytes to write */
+	master_key_info = getMasterKey(MasterKeyName, true, true);
 
 	memset(&int_key, 0, sizeof(InternalKey));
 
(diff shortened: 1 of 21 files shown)


src/encryption/enc_aes.c
```diff
@@ -151,11 +164,61 @@ void AesEncrypt(const unsigned char* key, const unsigned char* iv, const unsigne
...
+	AesEncrypt(master_key_info->data.data, iv, ((unsigned char*)rel_key_data) + SizeOfRelKeysDataHeader, INTERNAL_KEY_LEN, ((unsigned char *)(*p_enc_rel_key_data)) + SizeOfRelKeysDataHeader, (int *)enc_key_bytes);
+}
+
 void AesDecrypt(const unsigned char* key, const unsigned char* iv, const unsigned char* in, int in_len, unsigned char* out, int* out_len)
 {
 	AesRun(0, key, iv, in, in_len, out, out_len);
 }
 
+/*
+ * Provide a simple interface to decrypt a given key.
```
dutow, line 197: The comment specifies one key, but this can return multiple internal keys in one batch.

src/encryption/enc_aes.c
```diff
@@ -151,11 +164,61 @@ void AesEncrypt(const unsigned char* key, const unsigned char* iv, const unsigne
...
+void
+AesEncryptKey(const keyInfo *master_key_info, RelKeysData *rel_key_data, RelKeysData **p_enc_rel_key_data, size_t *enc_key_bytes)
+{
+	size_t sz;
+	SET_IV(iv);
+
+	/* Ensure we are getting a valid pointer here */
+	Assert(master_key_info);
+
+	sz = SizeOfRelKeysData(1);
```
dutow, line 183: How do we know that rel_key_data only contains one key?

src/encryption/enc_aes.c
```diff
@@ -151,11 +164,61 @@ void AesEncrypt(const unsigned char* key, const unsigned char* iv, const unsigne
...
+
+	*p_enc_rel_key_data = (RelKeysData *) palloc(sz);
+	memcpy(*p_enc_rel_key_data, rel_key_data, sz);
```
dutow, line 186: Inconsistent naming: `key` vs `keys`

src/access/pg_tde_tdemap.c
```diff
@@ -19,54 +19,97 @@
...
+
+#define MAP_ENTRY_FREE					0x00
+#define MAP_ENTRY_VALID					0x01
+
+#define MAP_ENTRY_SIZE					sizeof(TDEMapEntry)
+#define TDE_FILE_HEADER_SIZE			sizeof(TDEFileHeader)
+
+typedef struct TDEFileHeader
+{
+	int32 file_version;
```
dutow, line 61: Won't endianness be an issue with this?

src/access/pg_tde_tdemap.c
```diff
@@ -76,161 +119,19 @@ pg_tde_create_key_fork(const RelFileLocator *newrlocator, Relation rel)
...
-		ereport(DEBUG2,
-			(errmsg("fork file keys: [%lu] %s: %s", i+1, keys->master_key_name, tde_sprint_key(&keys->internal_key[i]))));
-#endif
-
-	pfree(key_file_path);
-	/* For now just close the key file.*/
-	FileClose(file);
-
-	return keys;
+	pg_tde_write_key_map_entry(newrlocator, &int_key, MasterKeyName);
```
dutow, line 134: This is incorrect, `MasterKeyName` is the generic name (`master-key`), not a name with the version included (`master-key-X`)

src/access/pg_tde_tdemap.c
```diff
@@ -250,19 +151,13 @@ GetRelationKeys(RelFileLocator rel)
...
-							rel_id)));
-#endif
+		if (curr->rel_id == rel_id)
+		{
 			return curr->keys;
 		}
 	}
 
-	keys = pg_tde_get_keys_from_fork(&rel);
+	keys = pg_tde_get_key_from_file(&rel, MasterKeyName);
```
dutow, line 160: Same issue: master key should be versioned, otherwise we can't rotate it.

src/access/pg_tde_tdemap.c
```diff
@@ -311,48 +206,652 @@ tde_sprint_masterkey(const keyData *k)
...
+void
+pg_tde_write_key_map_entry(const RelFileLocator *rlocator, InternalKey *key, const char *master_key_name)
+{
+	int32 key_index = 0;
+	const keyInfo *master_key_info;
+	RelKeysData *rel_key_data;
+	RelKeysData *enc_rel_key_data;
+
+	/* Get/generate a master, create the key for relation and get the encrypted key with bytes to write */
+	master_key_info = getMasterKey(master_key_name, true, true);
```
dutow, line 701: And again similar comment: This seems incorrect: it always gets the latest master key, even if the file already exists, and is encrypted with an older key. This here should load the key version specified at the beginning of the file (e.g. master-key-X), and use that for encrypting the new entry. Only if the file doesn't exists yet should it try to load the latest master key based on the generic name.

dutow (changes requested):
I added some comments.

My main issue is: the master key name handling seems to be completely incorrect.

Generally, we have two names:

* A generic name, `master-key`
* A specific version, `master-key-X`, e.g. `master-key-13`

An entire TDE file always has to be encrypted entirely with the same key version, otherwise we won't be able to decrypt it correctly.

This means that:
* If the TDE file doesn't exists, we have to use the generic name (`master-key`), lookup the latest master key version `master-key-X` using `getLatestKey` / `generateKey`, and then save the name of the specific key it gives us (with the version number) to the file.
* When adding a key to an existing file, we have to lookup the specific master key name stored in the file (`master-key-X`), use `keyringGetKey` to retrieve it, and encrypt the new internal key using that specific master key
* When reading an existing file, similarly we have to use `keyringGetKey` to retrieve the specific version specified in the file

src/encryption/enc_aes.c
```diff
@@ -151,11 +164,61 @@ void AesEncrypt(const unsigned char* key, const unsigned char* iv, const unsigne
...
 
+/*
+ * Provide a simple interface to encrypt a given key.
+ *
+ * The function pallocs and updates the p_enc_rel_key_data along with key bytes. The memory
+ * is allocated in the current memory context as this key should be ephemeral with a very
+ * short lifespan until it is written to disk.
+ */
+void
+AesEncryptKey(const keyInfo *master_key_info, RelKeysData *rel_key_data, RelKeysData **p_enc_rel_key_data, size_t *enc_key_bytes)
```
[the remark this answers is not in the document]
EngineeredVirus: It's essentially our interface to the SSL functions. Hence, I put these two functions here.

src/access/pg_tde_tdemap.c
EngineeredVirus: What challenge do you see here?

src/encryption/enc_aes.c
```diff
@@ -43,6 +46,16 @@
...
  */
 
+/*
+ * Let's have a single definition of the IV across the code. SET_IV should
+ * be at the end of a variable declaration block. If not, a compilation
+ * warning will be generated informing of nonconformance ISO C90 due to
+ * mixing of declarations and code.
+ */
+typedef unsigned char TDE_iv[INTERNAL_KEY_LEN];
+#define SET_IV(iv)                                      \
```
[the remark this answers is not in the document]
EngineeredVirus: I've removed it in favour of a bigger refactoring of this code.

src/encryption/enc_aes.c
EngineeredVirus: This will be fixed as we refactor the keyring API code.

src/encryption/enc_aes.c
EngineeredVirus: This will be fixed as we refactor the keyring API code.

src/encryption/enc_aes.c
EngineeredVirus: This will be fixed as we refactor the keyring API code.

src/access/pg_tde_tdemap.c
```diff
@@ -76,161 +119,19 @@ pg_tde_create_key_fork(const RelFileLocator *newrlocator, Relation rel)
...
-
-	mkey_len = (uint8) strlen(master_key_info->name.name);
 	/* XLOG internal keys */
 	XLogBeginInsert();
 	XLogRegisterData((char *) newrlocator, sizeof(RelFileLocator));
-	XLogRegisterData((char *) &mkey_len, sizeof(uint8));
-	XLogRegisterData((char *) master_key_info->name.name, strlen(master_key_info->name.name)+1);
-	XLogRegisterData((char *) &int_key_enc, sizeof(InternalKey));
+	XLogRegisterData((char *) &int_key, sizeof(InternalKey));
 	XLogInsert(RM_TDERMGR_ID, XLOG_TDE_CREATE_FORK);
```
dutow, line 126: `CREATE_FORK` is no longer a valid name, this should besomething else.

src/access/pg_tde_tdemap.c
```diff
@@ -76,161 +119,19 @@ pg_tde_create_key_fork(const RelFileLocator *newrlocator, Relation rel)
...
-	AesEncrypt(master_key_info->data.data, iv, (unsigned char*) &int_key, INTERNAL_KEY_LEN, (unsigned char*) &int_key_enc, &encsz);
-
-	mkey_len = (uint8) strlen(master_key_info->name.name);
 	/* XLOG internal keys */
 	XLogBeginInsert();
 	XLogRegisterData((char *) newrlocator, sizeof(RelFileLocator));
-	XLogRegisterData((char *) &mkey_len, sizeof(uint8));
-	XLogRegisterData((char *) master_key_info->name.name, strlen(master_key_info->name.name)+1);
-	XLogRegisterData((char *) &int_key_enc, sizeof(InternalKey));
+	XLogRegisterData((char *) &int_key, sizeof(InternalKey));
```
dutow, line 125: Now we are writing unencrypted internal key to the XLog. Previously it was an encrypted internal key. This should stay encrypted.

src/access/pg_tde_tdemap.c
```diff
@@ -76,161 +119,19 @@ pg_tde_create_key_fork(const RelFileLocator *newrlocator, Relation rel)
...
 	XLogInsert(RM_TDERMGR_ID, XLOG_TDE_CREATE_FORK);
 
-	/* TODO: should DB crash after sending XLog, secondaries would create a fork
+	/*
+	 * TODO: should DB crash after sending XLog, secondaries would create a fork
```
dutow, line 129: TODO no longer valid: we don't create fork files.

src/access/pg_tde_tdemap.c
```diff
@@ -311,48 +206,652 @@ tde_sprint_masterkey(const keyData *k)
...
-     * .tde postfix to the file name
-     */
-    rel_file_path = relpathperm(*newrlocator, MAIN_FORKNUM);
-    if (rel_file_path)
-    {
-        key_file_path = psprintf("%s."TDE_FORK_EXT, rel_file_path);
-        pfree(rel_file_path);
-    }
-    return key_file_path;
+	RelKeysData 	*rel_key_data;
```
dutow, line 216: Typo in name, should be `keys`

src/access/pg_tde_tdemap.c
```diff
@@ -311,48 +206,652 @@ tde_sprint_masterkey(const keyData *k)
...
+	return rel_key_data;
 }
 
-void 
-pg_tde_decrypt_internal_key(const InternalKey *in, InternalKey *out, const char *masterkeyname)
+/*
+ * Encrypts a given key and returns the encrypted one.
+ */
+RelKeysData *
+tde_encrypt_rel_key(const keyInfo *master_key_info, RelKeysData *rel_key_data)
```
dutow, line 234: Type in name, should be `keys`

src/access/pg_tde_tdemap.c
```diff
@@ -311,48 +206,652 @@ tde_sprint_masterkey(const keyData *k)
...
+	AesEncryptKey(master_key_info, rel_key_data, &enc_rel_key_data, &enc_key_bytes);
+
+	return enc_rel_key_data;
+}
+
+/*
+ * Decrypts a given key and returns the decrypted one.
+ */
+RelKeysData *
+tde_decrypt_rel_key(const keyInfo *master_key_info, RelKeysData *enc_rel_key_data)
```
dutow, line 248: Typo in name, should be `keys`

src/access/pg_tde_tdemap.c
```diff
@@ -311,48 +206,652 @@ tde_sprint_masterkey(const keyData *k)
...
+	snprintf(db_keydata_path, MAXPGPATH, "%s/%s%s", db_path, PG_TDE_KEYDATA_FILENAME, ((str_append) ? str_append : ""));
+}
+
+/*
+ * Path data clean up once the transaction is done.
+ */
+void
+pg_tde_cleanup_path_vars(void)
+{
+	*db_path = *db_map_path = *db_keydata_path = 0;
```
dutow, line 283: Should be `NULL`

src/encryption/enc_aes.c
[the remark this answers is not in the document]
dutow: No, this function already uses the functions defined in `enc_aes.h`. The idea is that `enc_aes` uses OpenSSL directly, and doesn't depend on postgres/pg_tde headers -  see the includes above, it had no entries related to these, before this PR.

Anything that depends on postgres/pg_tde headers is implemented in `enc_tuple.h`.

src/access/pg_tde_tdemap.c
EngineeredVirus: Agreed.

src/access/pg_tde_tdemap.c
EngineeredVirus: Resolved per my understanding.

src/access/pg_tde_tdemap.c
EngineeredVirus: Resolved.

src/access/pg_tde_tdemap.c
EngineeredVirus: Resolved.

src/access/pg_tde_tdemap.c
EngineeredVirus: Resolved.

src/access/pg_tde_tdemap.c
EngineeredVirus: This should be resolved when we remove the flexible array from the RelKeysData data structure.

src/access/pg_tde_tdemap.c
EngineeredVirus: This should be resolved when we remove the flexible array from the RelKeysData data structure.

src/access/pg_tde_tdemap.c
EngineeredVirus: This should be resolved when we remove the flexible array from the RelKeysData data structure.

src/encryption/enc_aes.c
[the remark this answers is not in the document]
EngineeredVirus: Resolved.

dutow (approved):

src/access/pg_tde_tdemap.c
```diff
@@ -19,218 +19,127 @@
...
+	XLogBeginInsert();
+	XLogRegisterData((char *) newrlocator, sizeof(RelFileLocator));
+	XLogRegisterData((char *) enc_rel_key_data->internal_key, sizeof(InternalKey));
+	XLogInsert(RM_TDERMGR_ID, XLOG_TDE_RELATION_KEY);
 
-	return keys;
+	/*
+	 * Add the encyrpted key to the key map data file structure.
+	 */
+	pg_tde_write_key_map_entry(newrlocator, enc_rel_key_data, master_key_info);
```
dAdAbird, line 142: Better to add a new internal key to the memory cache as well, as there is a big chances that the key will be used (no INSERT) right after the relation creation. This is how it was before and I think it makes sense from the performance perspective

src/keyring/keyring_api.c
```diff
@@ -171,3 +171,31 @@ const keyInfo* keyringGenerateKey(const char* internalName, unsigned keyLen)
...
+/*
+ * Simplifying the interface to get the master key without having to worry
+ * generating a new one. If master key does not exist, and doGenerateKey is
+ * set, a new key is generated. This is useful during write operations.
+ *
+ * However, when performing a read operation and a master is expected to exist,
+ * doGenerateKey should be false and doRaiseError should be set to indicate
+ * that master key is expected but could not be accessed.
+ */
+const keyInfo* getMasterKey(const char* internalName, int doGenerateKey, int doRaiseError)
```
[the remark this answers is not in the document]
EngineeredVirus: Absolutely. 🤦

src/access/pg_tde_tdemap.c
EngineeredVirus: Resolved per my understanding.

src/access/pg_tde_tdemap.c
EngineeredVirus: These aren't pointers, so to indicate that, I prefer 0 instead of NULL.

EngineeredVirus: Resolved.

dAdAbird (approved):

PR #117, opened 2024-02-12
Improving performance

### LINKED ISSUE #89: Performance issue with JSONB and JSON fields
I ran some tests that I had prepared to compare the performance of different datatypes and saw that JSONB performance is around 2.5x lower and JSON 1.5x lower.

The code I used to run the test is available here: https://github.com/waltton/pg_json_bench/tree/pg_tde
(there is a branch with a couple of changes to support `pg_tde`)

It runs:
```SQL
SELECT COUNT(*) FROM %s WHERE CAST(data->>'score' AS FLOAT) > 7.0
```

The table that is used could have a column as JSON or JSONB, using heap or pg_tde as access method.

The data comes from https://raw.githubusercontent.com/algolia/datasets/master/movies/records.json and each item in the array goes to one row in the db.

image

dutow:
This commit tries to improve the performance of our encryption, based on the results reported in #89.

It contains two main changes which results in improvents:

* Our pg_tde_crypt loop has several conditions to support all possible cornercases it might encounter. These conditions are not neccessary most of the time, but slow down our throughput considerably. This commit introduces a "simple" version of the function for smaller data sizes, which only has a simple one instruction for loop.
* OpenSSL supports in place encryption when certain conditions are met - which is true for our zeroblocks function. The zero block function no longer uses an additional local array, and the filling of the input data is also simplified.

These changes together results in the following encryption overheads:

* json testcase: 140% -> 120%
* jsonb testcase: 230% -> 175%

Closes #89

### DIFF
--- a/src/encryption/enc_tde.c
+++ b/src/encryption/enc_tde.c
@@ -48,7 +48,7 @@ SetIVPrefix(ItemPointerData* ip, char* iv_prefix)
  * start_offset: is the absolute location of start of data in the file.
  * This function assumes that everything is in a single block, and has an assertion ensuring this
  */
-void
+static void
 pg_tde_crypt_simple(const char* iv_prefix, uint32 start_offset, const char* data, uint32 data_len, char* out, RelKeyData* key, const char* context)
 {
 	const uint64 aes_start_block = start_offset / AES_BLOCK_SIZE;
@@ -84,7 +84,7 @@ pg_tde_crypt_simple(const char* iv_prefix, uint32 start_offset, const char* data
  * start_offset: is the absolute location of start of data in the file.
  * This is a generic function indented for large data, that od not fit into a single block
  */
-void
+static void
 pg_tde_crypt_complex(const char* iv_prefix, uint32 start_offset, const char* data, uint32 data_len, char* out, RelKeyData* key, const char* context)
 {
 	const uint64 aes_start_block = start_offset / AES_BLOCK_SIZE;
--- a/src/encryption/enc_tde.c
+++ b/src/encryption/enc_tde.c
@@ -43,16 +43,53 @@ SetIVPrefix(ItemPointerData* ip, char* iv_prefix)
  */
 
 /* 
- * pg_tde_crypt:
+ * pg_tde_crypt_simple:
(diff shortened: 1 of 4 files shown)


src/include/encryption/enc_tuple.h
```diff
@@ -16,6 +16,10 @@
 #include "executor/tuptable.h"
 #include "access/pg_tde_tdemap.h"
 
+extern void
+pg_tde_crypt_simple(const char* iv_prefix, uint32 start_offset, const char* data, uint32 data_len, char* out, RelKeysData* keys, const char* context);
+extern void
+pg_tde_crypt_complex(const char* iv_prefix, uint32 start_offset, const char* data, uint32 data_len, char* out, RelKeysData* keys, const char* context);
```
dAdAbird, line 22: Just a minor thing - should those be "external" (I mean to it looks like rather "private" functions)

dAdAbird (approved):

src/include/encryption/enc_tuple.h
dutow: Originally I considered simply calling the simple function at some places, but it is safer to always call the main function and it doesn't matter that much. I'll remove these from the header.

src/encryption/enc_tde.c
```diff
@@ -43,16 +43,53 @@ SetIVPrefix(ItemPointerData* ip, char* iv_prefix)
...
+	Aes128EncryptedZeroBlocks(&(key->internal_key.ctx), key->internal_key.key, iv_prefix, aes_start_block, aes_end_block, enc_key);
+
+#ifdef ENCRYPTION_DEBUG
 {
-	uint64 aes_start_block = start_offset / AES_BLOCK_SIZE;
-	uint64 aes_end_block = (start_offset + data_len + (AES_BLOCK_SIZE -1)) / AES_BLOCK_SIZE;
-	uint64 aes_block_no = start_offset % AES_BLOCK_SIZE;
+	char ivp_debug[33];
+	iv_prefix_debug(iv_prefix, ivp_debug);
+	ereport(LOG,
```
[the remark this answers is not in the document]
dutow: It's inside an `ENCRYPTION_DEBUG` macro

codeforall (approved):

PR #119, opened 2024-02-19
Also use the CommandID in tuple IVs

### LINKED ISSUE #112: Repeated use of AES-CTR on the same block is vulnerable to a chosen-plaintext attack
AES-CTR, as with all stream ciphers (and block cipher modes that emulate a stream cipher) is vulnerable to key stream recovery if the same key stream is used to encrypt multiple blocks of data, when at least one of the encryptions is over a known plaintext (either predictable, such as all-zeroes, or provided by an attacker).

The vulnerability comes from the fact that the per-block key, which I'll call `X`, is statically derived from the encryption of the block counter with the main key.  The stored ciphertext is simply the XOR of the plaintext, `P`, with the per-block key, `X`, to produce `C`.  If plaintext `P` is known, the per-block key `X` can be trivially recovered with `C XOR X`.  This allows all other encryptions of the same block to be decrypted with `C_n XOR X`.

The most appropriate solution to this problem would be to switch to using a tweakable cipher mode, such as AES-XTS, for which there is an existing OpenSSL function, [EVP_aes_128_xts](https://www.openssl.org/docs/man3.2/man3/EVP_aes_128_xts.html).

dutow:
Currently tupple iv calculatino is based only on the CTID, which means that when postgres reuses the same address, it also reuses the same IV. This makes our encryption theoretically weaker, in case of some unlikely but not impossible attack scenarios.

As an improvement this commit also adds the command id into the calculation. As the (ctid, cid) pair will be basically always unique, this solves the problem.

Fixes #112

### DIFF
--- a/src/encryption/enc_tde.c
+++ b/src/encryption/enc_tde.c
@@ -20,13 +20,16 @@ static void iv_prefix_debug(const char* iv_prefix, char* out_hex)
 #endif
 
 static void
-SetIVPrefix(ItemPointerData* ip, char* iv_prefix)
+SetIVPrefix(ItemPointerData* ip, CommandId cid, char* iv_prefix)
 {
 	/* We have up to 16 bytes for the entire IV
 	 * The higher bytes (starting with 15) are used for the incrementing counter
 	 * The lower bytes (in this case, 0..5) are used for the tuple identification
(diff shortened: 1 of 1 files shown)


dAdAbird (approved):

codeforall (approved):
LGTM

PR #127, opened 2024-02-28
Fix crash while deleting tuple w/ va attributes

### LINKED ISSUE #126: Crash while deleting tuple with varlen attribute

dAdAbird:
During the delete, we deform a tuple to create a new one with only Replica Identity attributes. Deforming - extracting attributes from the tuple, see `pg_tde_deform_tuple()` in `ExtractReplicaIdentity()` - needs to know the size of the variable length (va) attribute, which is in the va_header. As the va_header is the first N bytes of the actual data, the tuple has to be decrypted to read the attribute size.

Closes #126

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -13,7 +13,8 @@ non_sorted_off_compact \
 update_compare_indexes \
 pgtde_is_encrypted \
 multi_insert \
-trigger_on_view
+trigger_on_view \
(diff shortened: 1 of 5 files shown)


codeforall (approved):
Looks good

dutow (approved):

PR #132, opened 2024-03-01
Disallow deletion of keyring used by Master key.

codeforall:
The commit adds the before-delete trigger on the keyring catalog to ensure that the key provider used by the master key should not get deleted.

### DIFF
--- a/sql/keyprovider_dependency.sql
+++ b/sql/keyprovider_dependency.sql
@@ -1,0 +1,15 @@
+CREATE EXTENSION pg_tde;
+
+SELECT pg_tde_add_key_provider_file('mk-file-valut','/tmp/pg_tde_test_keyring.per');
+SELECT pg_tde_add_key_provider_file('free-file-valut','/tmp/pg_tde_test_keyring_2.per');
+SELECT pg_tde_add_key_provider_vault_v2('V2-Wallet','vault-token','percona.com/vault-v2/percona','/mount/dev','ca-cert-auth');
+
+SELECT pg_tde_set_master_key('test-db-master-key','mk-file-valut');
+
+-- Try dropping the in-use key provider
+DELETE FROM percona_tde.pg_tde_key_provider WHERE provider_name = 'mk-file-valut'; -- Should fail
+-- Now delete the un-used  key provider
+DELETE FROM percona_tde.pg_tde_key_provider WHERE provider_name = 'free-file-valut'; -- Should pass
+DELETE FROM percona_tde.pg_tde_key_provider WHERE provider_name = 'V2-Wallet'; -- Should pass
+
+DROP EXTENSION pg_tde;
--- a/sql/keyprovider_dependency.sql
+++ b/sql/keyprovider_dependency.sql
@@ -1,15 +1,15 @@
 CREATE EXTENSION pg_tde;
 
-SELECT pg_tde_add_key_provider_file('mk-file-valut','/tmp/pg_tde_test_keyring.per');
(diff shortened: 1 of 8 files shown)


sql/vault_v2_test.sql
```diff
@@ -1,7 +1,7 @@
 CREATE EXTENSION pg_tde;
 
 \getenv root_token ROOT_TOKEN
-SELECT pg_tde_add_key_provider_vault_v2('vault-v2',:'root_token','http://127.0.0.1:8200','secret',NULL);
+SELECT pg_tde_add_key_provider_vault_v2('vault-v2','root_token','http://127.0.0.1:8200','secret',NULL);
```
dutow, line 4: Why this change? `root_token` is a variable declared above, and it comes from an environment variable. This causes vault tests to fail on all CI runs

sql/keyprovider_dependency.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION pg_tde;
+
+SELECT pg_tde_add_key_provider_file('mk-file-valut','/tmp/pg_tde_test_keyring.per');
+SELECT pg_tde_add_key_provider_file('free-file-valut','/tmp/pg_tde_test_keyring_2.per');
+SELECT pg_tde_add_key_provider_vault_v2('V2-Wallet','vault-token','percona.com/vault-v2/percona','/mount/dev','ca-cert-auth');
+
+SELECT pg_tde_set_master_key('test-db-master-key','mk-file-valut');
+
+-- Try dropping the in-use key provider
+DELETE FROM percona_tde.pg_tde_key_provider WHERE provider_name = 'mk-file-valut'; -- Should fail
```
dutow, line 10: should updates fail too?

sql/keyprovider_dependency.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION pg_tde;
+
+SELECT pg_tde_add_key_provider_file('mk-file-valut','/tmp/pg_tde_test_keyring.per');
+SELECT pg_tde_add_key_provider_file('free-file-valut','/tmp/pg_tde_test_keyring_2.per');
+SELECT pg_tde_add_key_provider_vault_v2('V2-Wallet','vault-token','percona.com/vault-v2/percona','/mount/dev','ca-cert-auth');
```
dutow, line 5: please use the same environment based vault configuration as the existing vault test

sql/vault_v2_test.sql
codeforall: That was an oversight. Fixed!

sql/keyprovider_dependency.sql
codeforall: Not 100% sure about that. Let's discuss it and tackle it as a separate item if required.

sql/keyprovider_dependency.sql
```diff
@@ -0,0 +1,15 @@
+CREATE EXTENSION pg_tde;
+
+SELECT pg_tde_add_key_provider_file('mk-file-valut','/tmp/pg_tde_test_keyring.per');
```
dutow, line 3: There's a typo again: `valut`

Also, I agree with Andrew that we shouldn't call them vaults, as having a hashicorp vault provider, and calling file based keyrings vaults is confusing.

codeforall: Done!

dutow (approved):

dAdAbird (approved):

PR #135, opened 2024-03-04
Fix internal key XLogging for keyring changes

dAdAbird:
There is no functional relation cache during XLog replay therefore `GetMasterKey()` would fail looking for the keyring namespace. Actually, it would fail earlier on `get_master_key_info_path()`  as XLog replay has its own `MyDatabaseId` and `MyDatabaseTableSpace`. The latter could be fixed by providing `RelFileLocator` to `GetMasterKey()`  but the former can't be fixed easily (if it could be fixed at all) as XLog replay deals with the `Buffer` and shouldn't have access to the relations. To solve this, we should pass the master key name in the XLog so there is no need to call `GetMasterKey()` during XLog redo.

It also doesn't make sense to add an internal key to the cache during XLog replay as this cache is per-process and it won't be available to any user connections.

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -70,22 +70,27 @@ typedef struct TDEMapEntry
 	int32 key_index;
 } TDEMapEntry;
 
+typedef struct XLogRelKey
+{
+	RelFileLocator rlocator;
+	RelKeyData	relKey;
+} XLogRelKey;
+
 /* Global variables */
 static char db_path[MAXPGPATH] = {0};
 static char db_map_path[MAXPGPATH] = {0};
 static char db_keydata_path[MAXPGPATH] = {0};
 
 static void put_key_into_map(Oid rel_id, RelKeyData *key);
-static void pg_tde_xlog_create_relation(XLogReaderState *record);
 
-static RelKeyData* tde_create_rel_key(const RelFileLocator *rlocator, InternalKey *key, const TDEMasterKey *master_key, bool is_key_decrypted);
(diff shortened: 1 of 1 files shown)


src/access/pg_tde_tdemap.c
```diff
@@ -680,18 +686,18 @@ pg_tde_read_one_keydata(File keydata_file, off_t header_size, int32 key_index, c
...
+pg_tde_write_key_map_entry(const RelFileLocator *rlocator, RelKeyData *enc_rel_key_data)
 {
 	int32 key_index = 0;
 
 	/* Set the file paths */
 	pg_tde_set_db_file_paths(rlocator, NULL);
 
 	/* Create the map entry and then add the encrypted key to the data file */
-	key_index = pg_tde_write_map_entry(rlocator, db_map_path, master_key->keyName);
+	key_index = pg_tde_write_map_entry(rlocator, db_map_path, enc_rel_key_data->master_key_name);
```
dutow, line 697: This is not related to this PR, but I realized that we have an issue here: `pg_tde_write_map_entry` should receive the entire TDEMasterKey structure (with the version), not just the name (without the version).

src/access/pg_tde_tdemap.c
```diff
@@ -70,22 +70,27 @@ typedef struct TDEMapEntry
 	int32 key_index;
 } TDEMapEntry;
 
+typedef struct XLogRelKey
+{
+	RelFileLocator rlocator;
+	RelKeyData	relKey;
```
dutow, line 76: Generic question: can we agree on something and keep the naming conventions of structs the same? For example the struct above this uses under_score and camelCase both.

dAdAbird:
It was decided to make another PR w/ Xlogging of MasterKey data

PR #137, opened 2024-03-07
Fix internal key XLogging for keyring changes

dAdAbird:
Fix internal key XLogging for keyring changes 

1. There is no functional relation cache during XLog replay therefore
`GetMasterKey()` would fail looking for the keyring namespace. Actually,
it would fail earlier on `get_master_key_info_path()`  as XLog replay
has its own `MyDatabaseId` and `MyDatabaseTableSpace`. The latter could
be fixed by providing `RelFileLocator` to `GetMasterKey()`  but the former
can't be fixed easily (if it could be fixed at all) as XLog replay deals
with the `Buffer` and shouldn't have access to the relations. To solve this,
we should pass the master key name in the XLog so there is no need to call
`GetMasterKey()` during XLog redo.

It also doesn't make sense to add an internal key to the cache during XLog
replay as this cache is per-process and it won't be available to any user
connections.

2. Replicate the Master Key Info file on `pg_tde_set_master_key()`.

3. Use Master Key Version along with the Master Key Name in `pg_tde.*` headers.

### DIFF
--- a/src/catalog/tde_master_key.c
+++ b/src/catalog/tde_master_key.c
@@ -10,6 +10,8 @@
  *-------------------------------------------------------------------------
  */
 #include "postgres.h"
+#include "access/xlog.h"
+#include "access/xloginsert.h"
 #include "catalog/tde_master_key.h"
 #include "common/pg_tde_shmem.h"
 #include "storage/lwlock.h"
@@ -22,6 +24,7 @@
 #include "funcapi.h"
 #include "utils/builtins.h"
 #include "pg_tde.h"
+#include "access/pg_tde_xlog.h"
 #include <sys/time.h>
 
 #define PG_TDE_MASTER_KEY_FILENAME "tde_master_key.info"
@@ -55,7 +58,7 @@ static dshash_parameters master_key_dsh_params = {
 
 TdeMasterKeylocalState masterKeyLocalState;
 
-static char *get_master_key_info_path(void);
+static char *get_master_key_info_path(TDEMasterKeyInfo *masterKeyInfo);
 static void master_key_info_attach_shmem(void);
 static Size initialize_shared_state(void *start_address);
 static void initialize_objects_in_dsa_area(dsa_area *dsa, void *raw_dsa_area);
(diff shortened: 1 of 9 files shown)


src/catalog/tde_master_key.c
```diff
@@ -256,11 +267,6 @@ get_master_key_info(void)
 
     masterKeyInfo = palloc(sizeof(TDEMasterKeyInfo));
     bytes_read = FileRead(master_key_file, masterKeyInfo, sizeof(TDEMasterKeyInfo), 0, WAIT_EVENT_DATA_FILE_READ);
-    if (bytes_read == 0)
```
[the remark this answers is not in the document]
dAdAbird: Otherwise, we'd fail to write to it, as in `save_master_key_info()` we open it with `O_EXCL`. We can add checks there - rewrite the file if it exists and it's empty and return with an error if it is not empty. Or rewrite it in any case (which I don't think is a good idea). But all this will complicate the logic when I don't see the value of "allowing" empty files while switching the strategy to delete them on cleanups. I mean in what legitimate scenario there can be an empty master key info file?

src/include/access/pg_tde_xlog.h
```diff
@@ -0,0 +1,34 @@
...
+#define XLOG_TDE_CLEAN_MASTER_KEY	0x20
+/* TODO: ID has to be registedred and changed: https://wiki.postgresql.org/wiki/CustomWALResourceManagers */
+#define RM_TDERMGR_ID	RM_EXPERIMENTAL_ID
+#define RM_TDERMGR_NAME	"test_pg_tde_custom_rmgr"
+
+extern void pg_tde_rmgr_redo(XLogReaderState *record);
+extern void pg_tde_rmgr_desc(StringInfo buf, XLogReaderState *record);
+extern const char *pg_tde_rmgr_identify(uint8 info);
+
+/* Move this to pg_tde.c file */
```
codeforall, line 26: Why do we want to do that?

dAdAbird: This is the part of the resource manager interface (`RmgrData`) those are required functions. They are used by the backend for logging etc

EngineeredVirus (approved):
LGTM

PR #138, opened 2024-03-07
Remove remaining json-c dependency from keyring vault

dutow:
After this change, pg_tde no longer needs json-c as a dependency.


src/keyring/keyring_vault.c
```diff
@@ -128,30 +130,29 @@ set_key_by_name(GenericKeyring* keyring, keyInfo *key, bool throw_error)
...
 	Assert(key != NULL);
 
+	// Since we are only building a very limited JSON with a single base64 string, we build it by hand
+	// Simpler than using the limited pg json api
 	keyLen = pg_b64_encode((char *)key->data.data, key->data.len, keyData, 64);
 	keyData[keyLen] = 0;
-	json_object_object_add(data, "key", json_object_new_string(keyData));
-	json_object_object_add(request, "data", data);
+	
+	snprintf(jsonText, 512, "{\"data\":{\"key\":\"%s\"}}", keyData);
```
codeforall, line 144: We should make a function to output a JSON string as this gets repeated multiple times in the PR.

src/keyring/keyring_vault.c
```diff
@@ -210,64 +211,17 @@ get_key_by_name(GenericKeyring *keyring, const char *key_name, bool throw_error,
...
-		*return_code = KEYRING_CODE_INVALID_RESPONSE;
-		ereport(throw_error ? ERROR : WARNING,
-				(errmsg("keyring provider \"%s\" returned invalid JSON",
-						vault_keyring->keyring.provider_name),
-				 errdetail("Key doesn't exist or empty.")));
-		goto cleanup;
-	}
+	dataJson = DirectFunctionCall2(json_object_field_text, CStringGetTextDatum(str.ptr), CStringGetTextDatum("data"));
+	data2Json = DirectFunctionCall2(json_object_field_text, dataJson, CStringGetTextDatum("data"));
+	keyJson = DirectFunctionCall2(json_object_field_text, data2Json, CStringGetTextDatum("key"));
```
codeforall, line 216: Is there any possibility of these Direct function calls returning NULL? If so we need to encapsulate these in PG_TRY PG_CATCH block

src/keyring/keyring_vault.c
dutow: This is the only place where we create a json object.

src/keyring/keyring_vault.c
dutow: Fixed

codeforall (approved):

PR #141, opened 2024-03-13
Refactoring the code and adding TDEMasterKeyInfo to TDE file headers.

EngineeredVirus:
This also eliminates the master key info file in favour of saving the same information in the TDE map and key data files.

This is the first step towards master key rotation as we are eliminating the need for making changes in multiple files and thereby, making the master key rotation an atomic operation.

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -23,6 +23,8 @@
 #include "miscadmin.h"
 
 #include "access/pg_tde_tdemap.h"
+#include "access/pg_tde_xlog.h"
+#include "catalog/tde_master_key.h"
 #include "encryption/enc_aes.h"
 #include "encryption/enc_tde.h"
 #include "keyring/keyring_api.h"
@@ -60,7 +62,7 @@
 typedef struct TDEFileHeader
 {
 	int32 file_version;
-	char master_key_name[MASTER_KEY_NAME_LEN];
+	TDEMasterKeyId master_key_id;
 } TDEFileHeader;
 
 typedef struct TDEMapEntry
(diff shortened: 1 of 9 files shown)


src/access/pg_tde_tdemap.c
```diff
@@ -78,25 +79,29 @@ static char db_keydata_path[MAXPGPATH] = {0};
...
 
-static void pg_tde_set_db_file_paths(const RelFileLocator *rlocator, char *str_append);
-static File pg_tde_open_file(char *tde_filename, const char *master_key_name, int fileFlags, bool *is_new_file, off_t *offset);
-static void pg_tde_write_key_map_entry(const RelFileLocator *rlocator, RelKeyData *enc_rel_key_data, const TDEMasterKey *master_key);
+static void pg_tde_set_db_file_paths(void);
+static File pg_tde_open_file(char *tde_filename, TDEMasterKeyInfo *master_key_info, bool should_fill_info, int fileFlags, bool *is_new_file, off_t *offset);
+static void pg_tde_write_key_map_entry(const RelFileLocator *rlocator, RelKeyData *enc_rel_key_data, TDEMasterKey *master_key);
 
-static int32 pg_tde_write_map_entry(const RelFileLocator *rlocator, char *db_map_path, const char *master_key_name);
+static int32 pg_tde_write_map_entry(const RelFileLocator *rlocator, char *db_map_path, TDEMasterKey *master_key);
```
dAdAbird, line 95: I feel like `TDEMasterKeyInfo` is more appropriate here and in the rest of the functions in this file as we dealing with the key info only. It would be more clear for the code readers.

src/access/pg_tde_tdemap.c
```diff
@@ -283,80 +288,217 @@ pg_tde_cleanup_path_vars(void)
...
+ */
+void
+pg_tde_delete_tde_files(void)
+{
+	/* Set the file paths */
+	pg_tde_set_db_file_paths();
+
+	/* Open and vaidate file for basic correctness. */
+	PathNameDeleteTemporaryFile(db_map_path, true);
+	PathNameDeleteTemporaryFile(db_keydata_path, true);
```
dAdAbird, line 302: Won't `durable_unlink()`  be better here? "TemporaryFile" a bit looks confusing in this context plus `durable_unlink()` ensures the remove effect is persist before returning

PR #143, opened 2024-03-15
Refactoring the code and adding TDEMasterKeyInfo to TDE file headers.

EngineeredVirus:
This also eliminates the master key info file in favour of saving the same information in the TDE map and key data files.

This is the first step towards master key rotation as we are eliminating the need for making changes in multiple files and thereby, making the master key rotation an atomic operation.


src/access/pg_tde_tdemap.c
```diff
@@ -678,18 +830,20 @@ pg_tde_read_one_keydata(File keydata_file, off_t header_size, int32 key_index, c
...
 
+	Assert(rlocator);
+
 	/* Set the file paths */
-	pg_tde_set_db_file_paths(rlocator, NULL);
+	pg_tde_set_db_file_paths(rlocator->dbOid);
 
 	/* Create the map entry and then add the encrypted key to the data file */
-	key_index = pg_tde_write_map_entry(rlocator, db_map_path, &enc_rel_key_data->master_key_id, fail_on_check);
+	key_index = pg_tde_write_map_entry(rlocator, db_map_path, &master_key->keyInfo);
```
dAdAbird, line 843: [src/access/pg_tde_xlog.c](https://github.com/Percona-Lab/pg_tde/blob/81cb2b0647307463638c5a6a8487894fba4d1899/src/access/pg_tde_xlog.c#L35) still calls `pg_tde_write_key_map_entry()` with the `bool` argument. It is confusing at least  and I'm not sure how safe it is to dereference `keyInfo` here w/o a NULL check.

EngineeredVirus: Good catch. The compiler never complained, and I had hoped that changing function arguments would cause that. In this case, it didn’t because it probably took false as NULL.

codeforall (approved):

PR #144, opened 2024-03-15
Keyrotation and key name versioning implementation

EngineeredVirus:
The initial version of the key rotation functionality without the xlog functionality.
    
This patch includes some minor refactoring as well as changes to the return type of the set key function, which now returns a boolean.

Master key versioning implemented by Usama. A new field of versioned_name is added.
    
Co-authored-by: Muhammad Usama <m.usama@gmail.com>

### DIFF
--- a/pg_tde--1.0.sql
+++ b/pg_tde--1.0.sql
@@ -81,12 +81,12 @@ RETURNS boolean
 AS $$ SELECT amname = 'pg_tde' FROM pg_class INNER JOIN pg_am ON pg_am.oid = pg_class.relam WHERE relname = table_name $$
 LANGUAGE SQL;
 
-CREATE FUNCTION pg_tde_rotate_key(new_master_key_name VARCHAR(255), new_provider_name VARCHAR(255))
+CREATE FUNCTION pg_tde_rotate_key(new_master_key_name VARCHAR(255) DEFAULT NULL, new_provider_name VARCHAR(255) DEFAULT NULL, ensure_new_key BOOLEAN DEFAULT TRUE)
 RETURNS boolean
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
-CREATE FUNCTION pg_tde_set_master_key(master_key_name VARCHAR(255), provider_name VARCHAR(255))
+CREATE FUNCTION pg_tde_set_master_key(master_key_name VARCHAR(255), provider_name VARCHAR(255), ensure_new_key BOOLEAN DEFAULT FALSE)
 RETURNS boolean
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
--- a/pg_tde--1.0.sql
+++ b/pg_tde--1.0.sql
@@ -81,13 +81,13 @@ RETURNS boolean
 AS $$ SELECT amname = 'pg_tde' FROM pg_class INNER JOIN pg_am ON pg_am.oid = pg_class.relam WHERE relname = table_name $$
 LANGUAGE SQL;
 
-CREATE FUNCTION pg_tde_rotate_key(key_name VARCHAR)
+CREATE FUNCTION pg_tde_rotate_key(new_master_key_name VARCHAR(255), new_provider_name VARCHAR(255))
 RETURNS boolean
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
 CREATE FUNCTION pg_tde_set_master_key(master_key_name VARCHAR(255), provider_name VARCHAR(255))
-RETURNS VOID
+RETURNS boolean
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
 
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -979,12 +979,12 @@ void
 finalize_key_rotation(char *m_path_old, char *k_path_old, char *m_path_new, char *k_path_new)
 {
 	/* Remove old files */
-	durable_unlink(m_path_old, LOG);
-	durable_unlink(k_path_old, LOG);
+	durable_unlink(m_path_old, ERROR);
+	durable_unlink(k_path_old, ERROR);
 
 	/* Rename the new files to required filenames */
-	durable_rename(m_path_new, m_path_old, LOG);
-	durable_rename(k_path_new, k_path_old, LOG);
+	durable_rename(m_path_new, m_path_old, ERROR);
+	durable_rename(k_path_new, k_path_old, ERROR);
 }
 
 /*
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -102,6 +102,9 @@ static RelKeyData* pg_tde_get_key_from_file(const RelFileLocator *rlocator);
 static RelKeyData* pg_tde_read_keydata(char *db_keydata_path, int32 key_index, TDEMasterKey *master_key);
 static RelKeyData* pg_tde_read_one_keydata(File keydata_file, int32 key_index, TDEMasterKey *master_key);
 
+static File keyrotation_init_file(TDEMasterKeyInfo *new_master_key_info, char *rotated_filename, char *filename, bool *is_new_file, off_t *curr_pos);
+static void finalize_key_rotation(char *m_path_old, char *k_path_old, char *m_path_new, char *k_path_new);
+
 /*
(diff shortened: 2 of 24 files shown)


src/catalog/tde_master_key.c
```diff
@@ -373,18 +373,83 @@ set_master_key_with_keyring(const char *key_name, GenericKeyring *keyring)
...
+
+    /* We need a valid keyring structure */
+    keyring = GetKeyProviderByID(new_master_key.keyInfo.keyringId);
+
+
+    /* We need to get the key from keyring */
+    while (true)
+    {
+        keyInfo = KeyringGetKey(keyring, new_master_key.keyInfo.keyId.versioned_name, false, &keyring_ret);
+        if (keyInfo == NULL || ensure_new_key == false)
```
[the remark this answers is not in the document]
dAdAbird: BTW, should we always generate a new key during rotation? Can't a user set a key with the "new name" which is already exists in a keyring and we use it? And generate a new key only if a provided new name doesn't exists or the name is the same as the name of the current key (is such a case we generate a new version of the key). I'm just asking because I'm not sure what is the supposed logic in that regard.
Can we have a comment w/ these rules?

src/access/pg_tde_tdemap.c
```diff
@@ -949,45 +955,206 @@ pg_tde_get_key_from_file(const RelFileLocator *rlocator)
...
 
 /*
- * TODO:
- * 		- How do we get the old key name and the key itself?
- * 		- We need to specify this for a current or all databases?
+ * Rotate keys and generates the WAL record for it.
  */
 bool
-pg_tde_perform_rotate_key(const char *new_master_key_name)
+pg_tde_perform_rotate_key(TDEMasterKey *master_key, TDEMasterKey *new_master_key)
```
dAdAbird, line 994: Do we need locks in this function? Or should the caller have locks?
I imagine it can be call by different processes which can rewrite/unlink files at the same time.

src/catalog/tde_master_key.c
```diff
@@ -326,14 +307,34 @@ set_master_key_with_keyring(const char *key_name, GenericKeyring *keyring)
...
         /* We need to get the key from keyring */
+        while (true)
+        {
+            keyInfo = KeyringGetKey(keyring, masterKey->keyInfo.keyId.versioned_name, false, &keyring_ret);
+            if (keyInfo == NULL || ensure_new_key == false)
+                break;
+
+            masterKey->keyInfo.keyId.version++;
+
+            snprintf(masterKey->keyInfo.keyId.versioned_name, TDE_KEY_NAME_LEN, "%s_%d", key_name, masterKey->keyInfo.keyId.version);
```
dutow, line 325: Setting this variable here seems out of place - shouldn't it be part of keyring internals?

src/catalog/tde_master_key.c
```diff
@@ -326,14 +307,34 @@ set_master_key_with_keyring(const char *key_name, GenericKeyring *keyring)
...
         masterKey->keyInfo.keyringId = keyring->key_id;
         strncpy(masterKey->keyInfo.keyId.name, key_name, TDE_KEY_NAME_LEN);
+        strncpy(masterKey->keyInfo.keyId.versioned_name, key_name, TDE_KEY_NAME_LEN);
+        gettimeofday(&masterKey->keyInfo.creationTime, NULL);
+
         /* We need to get the key from keyring */
+        while (true)
+        {
+            keyInfo = KeyringGetKey(keyring, masterKey->keyInfo.keyId.versioned_name, false, &keyring_ret);
+            if (keyInfo == NULL || ensure_new_key == false)
```
dutow, line 320: This condition seems strange to me - if ensure_new_key is false we  always return the first version of the key, not the last?

src/catalog/tde_master_key.c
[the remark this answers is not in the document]
dutow: We should have functionality to use an existing key, for example when multiple databases use the same master key. One rotation should generate a new version, the others just should use the latest.

src/catalog/tde_master_key.c
```diff
@@ -326,14 +307,34 @@ set_master_key_with_keyring(const char *key_name, GenericKeyring *keyring)
...
+        masterKey->keyInfo.keyId.version = DEFAULT_MASTER_KEY_VERSION;
         masterKey->keyInfo.keyringId = keyring->key_id;
         strncpy(masterKey->keyInfo.keyId.name, key_name, TDE_KEY_NAME_LEN);
+        strncpy(masterKey->keyInfo.keyId.versioned_name, key_name, TDE_KEY_NAME_LEN);
+        gettimeofday(&masterKey->keyInfo.creationTime, NULL);
+
         /* We need to get the key from keyring */
+        while (true)
+        {
+            keyInfo = KeyringGetKey(keyring, masterKey->keyInfo.keyId.versioned_name, false, &keyring_ret);
```
dutow, line 319: I see now why you introduced versioned_name, but this is a hack and not a good solution. Instead of this, KeyringGetKey should receive keyInfo as a parameter, and deal with the name construction internally.

pg_tde--1.0.sql
```diff
@@ -81,13 +81,13 @@ RETURNS boolean
 AS $$ SELECT amname = 'pg_tde' FROM pg_class INNER JOIN pg_am ON pg_am.oid = pg_class.relam WHERE relname = table_name $$
 LANGUAGE SQL;
 
-CREATE FUNCTION pg_tde_rotate_key(key_name VARCHAR)
+CREATE FUNCTION pg_tde_rotate_key(new_master_key_name VARCHAR(255), new_provider_name VARCHAR(255))
```
dutow, line 84: 99% of the time, users just want to rotate the key, and that's it. We should have an overload without any arguments.

We also need a function that just updates the file to the latest already existing key, instead of creating a new one.

src/catalog/tde_master_key.c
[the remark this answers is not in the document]
codeforall: Basically, the loop appears redundant because of the missing SQL interface. The idea is that if users want to use the existing key ( for whatever reason), they call the set key function with mastrer-key name (using ensure_new_key=false) that already exists in the key provider. Secondly, if the user intends to use the latest version of the key when the key name already exists in the key provider, then they can do so by passing ensure_new_key=true.
This PR just adds the infrastructure for enabling the above.

src/catalog/tde_master_key.c
codeforall: IMHO the key version is the function of master_key, not the keyring. And keyring should only be concerned about key-name and if that key-name contains the version number appended to it so be it. Also managing the version as part of master key naming allows the keyring to get-set master key without constructing the key name every time and that is good for performance as well.

src/access/pg_tde_tdemap.c
```diff
@@ -953,9 +956,39 @@ pg_tde_get_key_from_file(const RelFileLocator *rlocator)
...
+void
+finalize_key_rotation(char *m_path_old, char *k_path_old, char *m_path_new, char *k_path_new)
+{
+	/* Remove old files */
+	durable_unlink(m_path_old, LOG);
+	durable_unlink(k_path_old, LOG);
+
+	/* Rename the new files to required filenames */
+	durable_rename(m_path_new, m_path_old, LOG);
+	durable_rename(k_path_new, k_path_old, LOG);
```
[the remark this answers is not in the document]
EngineeredVirus: Changing the elevel to ERROR for both primary and standby.

src/catalog/tde_master_key.c
```diff
@@ -373,18 +373,83 @@ set_master_key_with_keyring(const char *key_name, GenericKeyring *keyring)
...
+            break;
+
+        new_master_key.keyInfo.keyId.version++;
+
+        snprintf(new_master_key.keyInfo.keyId.versioned_name, TDE_KEY_NAME_LEN, "%s_%d", new_key_name, new_master_key.keyInfo.keyId.version);
+
+        if (new_master_key.keyInfo.keyId.version > MAX_MASTER_KEY_VERSION_NUM)
+        {
+            ereport(ERROR,
+                    (errmsg("Failed to generate new key name")));
```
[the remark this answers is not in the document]
EngineeredVirus: Resolved.

EngineeredVirus:
As agreed during the weekly meeting, the discussion points are now pretty much resolved by the latest commits from @codeforall and @EngineeredVirus, barring the locking issue, which will come as a separate PR for the whole structure, not just key rotation.

PR #146, opened 2024-03-20
Tech-preview-release-notes

nastena1606:


documentation/docs/release-notes/tech-preview.md
capri1989, line 1: Given that this is the first "release", I would like to have a full-blown feature document and not only enhancements and new features.

capri1989: or do you think this should be put into a blog and part of the final GA release?

documentation/docs/release-notes/release-notes.md
```diff
@@ -0,0 +1,45 @@
...
+* Fixed a bug where `pg_tde` sent many 404 requests to the Vault server
+* Fixed сompatibility issues with old OpenSSL versions
+* Fixed сompatibility with old Curl versions 
+
+## MVP (2023-12-12)
+
+The Mininum Vial Product (MVP) version introduces the following functionality:
+
+* Encryption of heap tables, including TOAST
+* Encryption key are stored either in Hashicorp Vault server or in local keyring file (for development)
```
capri1989, line 43: "Encryption keys" instead of key

capri1989 (approved):
LGTM

codeforall:
We decided to use Alpha1 as a release version instead of TechPreview. We need to reflect this in the release notes update.

documentation/docs/release-notes/release-notes.md
```diff
@@ -0,0 +1,45 @@
+# pg_tde release notes
+
+ ## Technical preview (2024-03-28)
```
capri1989, line 3: Should this here also state Alpha1 now?

codeforall (approved):

dutow (approved):

PR #148, opened 2024-03-21
Fix for Issue:140

codeforall:
Update the pgtde_is_encrypted() definition to return false when the table name does not exist.


dutow (approved):

EngineeredVirus (approved):
LGTM

PR #156, opened 2024-03-28
Fix IPC::Run perl module name in CI

artemgavrilov:
```
Warning: Cannot install IPC::RUN, don't know what it is.
Try the command

    i /IPC::RUN/

to find objects with matching identifiers.
```

https://metacpan.org/pod/IPC::Run

Same for pg_stat_monitor: https://github.com/percona/pg_stat_monitor/pull/438


Naeem-Akhter (approved):
LGTM!

dutow (approved):

PR #158, opened 2024-03-28
Add pg_tde_version() function to get the module version of the extension

codeforall:
To make the version number configurable, the configure script outputs the PACKAGE VERSION to the config.h file that gets returned by the SQL interface. To bump the version, you only need to update the version number in the AC_INIT macro in configure.ac file


src/include/pg_tde_defs.h
```diff
@@ -0,0 +1,16 @@
...
+ * src/include/pg_tde_defs.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef PG_TDE_DEFS_H
+#define PG_TDE_DEFS_H
+
+
+extern const char *pg_tde_package_string(void);
+extern const char * pg_tde_package_name(void);
```
dAdAbird, line 13:
```suggestion
extern const char *pg_tde_package_name(void);
```

just for the formatting consistency

dAdAbird (approved):

PR #183, opened 2024-04-24
XLog encryption prototype

dAdAbird:
Encrypts XLog pages while writing them to the segment file.
This is just a POC without any proper key management.


codeforall (approved):
We agreed to discuss the test case failures separately. Otherwise, the code changes look good.

PR #189, opened 2024-05-13
Implement event triggers to identify index creation on encrypted tables.

codeforall:
This commit implements ddl-start and ddl-end event triggers to identify index creation operations on encrypted tables. Upon creating an index on an encrypted table, the trigger function updates the global state, which can be accessed by the storage manager (mgr) to decide if smgr_create needs to do encryption or not.

The start-ddl function analyzes the CREATE TABLE and CREATE INDEX statements and identifies if the table uses the pg_tde access method. When the table is created or the one on which the index is being created utilizes the pg_tde access method, the start-ddl trigger function populates relevant information about the encrypted table into a global structure. This structure can be accessed using the GetCurrentTdeCreateEvent() function. After the execution of the current DDL command finishes, the end-ddl function clears out this structure.

### DIFF
--- a/pg_tde--1.0.sql
+++ b/pg_tde--1.0.sql
@@ -131,3 +131,21 @@ COMMENT ON ACCESS METHOD pg_tde IS 'pg_tde table access method';
 
 -- Per database extension initialization
 SELECT pg_tde_extension_initialize();
+
+CREATE OR REPLACE FUNCTION ddl_command_start_capture()
+RETURNS event_trigger
+AS 'MODULE_PATHNAME'
+LANGUAGE C;
+
+CREATE OR REPLACE FUNCTION ddl_command_end_capture()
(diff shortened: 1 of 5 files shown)


pg_tde--1.0.sql
```diff
@@ -131,3 +131,21 @@ COMMENT ON ACCESS METHOD pg_tde IS 'pg_tde table access method';
 
 -- Per database extension initialization
 SELECT pg_tde_extension_initialize();
+
+CREATE OR REPLACE FUNCTION ddl_command_start_capture()
```
dutow, line 135: shouldn't we use the `pg_tde` prefix with these functions / triggers?

dutow (commented):
I only added one minor comment. Please also change the target branch of the PR to the  `smgr` branch

PR #197, opened 2024-05-27
TDE TupleTableSlot for storing decrypted tuple along with the buffer …

codeforall:
…tuple

Tuple data in the shared buffer is encrypted. To store the tuple in the tupleTableslot, the tuple data is decrypted into allocated memory. This memory needs to be properly cleaned up. However, with the existing BufferHeapTupleTableSlot, there is no way to free this memory until the end of the current query executor cycle.

To address this, the commit introduces TDEBufferHeapTupleTableSlot, a clone of BufferHeapTupleTableSlot that keeps a reference to the allocated decrypted tuple and frees it when the tuple slot is cleared. Most of the code is borrowed from the BufferHeapTupleTableSlot implementation, ensuring that TDEBufferHeapTupleTableSlot can be cast to BufferHeapTupleTableSlot

### DIFF
--- a/src/access/pg_tde_slot.c
+++ b/src/access/pg_tde_slot.c
@@ -1,0 +1,536 @@
+/*-------------------------------------------------------------------------
+ *
+ * pg_tdeam.c
+ *	  pg_tde TupleTableSlot implementation code
+ *
+ * Portions Copyright (c) 1996-2024, PostgreSQL Global Development Group
+ * Portions Copyright (c) 1994, Regents of the University of California
+ * Portions Copyright (c) 2024, Percona
+ *
+ *
+ * IDENTIFICATION
+ *	  contrib/pg_tde/access/pg_tde_slot.c
+ *
+ *
+ */
+#include "postgres.h"
+#include "access/pg_tde_slot.h"
+#include "access/heaptoast.h"
+#include "access/htup_details.h"
+#include "access/tupdesc_details.h"
+#include "catalog/pg_type.h"
+#include "funcapi.h"
+#include "nodes/nodeFuncs.h"
+#include "storage/bufmgr.h"
+#include "utils/builtins.h"
+#include "utils/expandeddatum.h"
+#include "utils/lsyscache.h"
+#include "utils/typcache.h"
+#include "encryption/enc_tde.h"
+
+/*
+ * TTSOpsTDEBufferHeapTuple is effectively the same as TTSOpsBufferHeapTuple slot.
+ * The only difference is that it keeps the reference of the decrypted tuple
+ * and free it during clear slot operation
+ */
+
+const TupleTableSlotOps TTSOpsTDEBufferHeapTuple;
+static pg_attribute_always_inline void slot_deform_heap_tuple(TupleTableSlot *slot, HeapTuple tuple, uint32 *offp,
+															  int natts);
+static inline void tts_buffer_heap_store_tuple(TupleTableSlot *slot,
+											   HeapTuple tuple,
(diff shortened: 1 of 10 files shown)


src/access/pg_tde_slot.c
```diff
@@ -0,0 +1,536 @@
...
+		elog(ERROR, "trying to store an on-disk heap tuple into wrong type of slot");
+
+    if (rel->rd_rel->relkind != RELKIND_TOASTVALUE)
+    {
+        TDEBufferHeapTupleTableSlot *bslot = (TDEBufferHeapTupleTableSlot *) slot;
+        RelKeyData *key = GetRelationKey(rel->rd_locator);
+		bslot->decrypted_tuple = heap_copytuple(tuple);
+        PG_TDE_DECRYPT_TUPLE_EX(tuple, bslot->decrypted_tuple, key, "ExecStoreBuffer");
+		/* TODO: revisit this */
+		tuple->t_data = bslot->decrypted_tuple->t_data;
```
dAdAbird, line 489: 2. Have you considered having `memcpy(tuple->t_data, decrypted_tuple->t_data, ...)` instead?

src/access/pg_tde_slot.c
```diff
@@ -0,0 +1,536 @@
...
+
+	if (unlikely(!TTS_IS_TDE_BUFFERTUPLE(slot)))
+		elog(ERROR, "trying to store an on-disk heap tuple into wrong type of slot");
+
+    if (rel->rd_rel->relkind != RELKIND_TOASTVALUE)
+    {
+        TDEBufferHeapTupleTableSlot *bslot = (TDEBufferHeapTupleTableSlot *) slot;
+        RelKeyData *key = GetRelationKey(rel->rd_locator);
+		bslot->decrypted_tuple = heap_copytuple(tuple);
+        PG_TDE_DECRYPT_TUPLE_EX(tuple, bslot->decrypted_tuple, key, "ExecStoreBuffer");
```
dAdAbird, line 487: 1. Have you considered decrypting a in-place?
```
PG_TDE_DECRYPT_TUPLE_EX(tuple, tuple, key, "ExecStoreBuffer");
```

dAdAbird (commented):
I'd run `pgindent` on newly created files

src/access/pg_tde_slot.c
codeforall: Decrypting in-place will actually change the data in the shared buffer. Next time around, we will have no way of knowing if the data for that particular tuple is already decrypted. Secondly, the decrypted data could also get flushed to the disk if the same buffer gets dirty.

src/access/pg_tde_slot.c
```diff
@@ -0,0 +1,536 @@
...
+tts_buffer_heap_release(TupleTableSlot *slot)
+{
+}
+
+static void
+tts_buffer_heap_clear(TupleTableSlot *slot)
+{
+	TDEBufferHeapTupleTableSlot *bslot = (TDEBufferHeapTupleTableSlot *) slot;
+
+    if (bslot->decrypted_tuple)
```
dutow, line 62: Can we add some markers around our changes, so in the future it will be easier to update the original code? Maybe with a dummy define and ifdefs, or just with comments.

codeforall: That is a good point, but that needs to be done across the code base to have one consistent view of things when we want to identify the changes. Having it in only one place would cause more confusion.
So I think we should take it up as a separate TODO

dAdAbird (approved):

dutow (approved):

PR #199, opened 2024-05-29
Minimal working prototype for SMGR + Event trigger + keyring encryption 

dutow:
* Introduces `pg_tde2` access method
* New access method uses the event trigger changes from #196
* Keys are now loaded from the keyring
* This requires changes to the map file / master key infrastructure
* This commit only modifies/fixes those as little as required for simplicity.
  More refactoring/changes coming in separate commits / PRs
* Removes reliance from MyDatabaseId, as things now have to work with multiple databases in the checkpointer
* Removes some error reports, where functions should work even without a configured keyring
* Fixes some bugs in the map file functions
* Map file functions now work with multiple databases in a single process, but this is a hackish solution, global state needs a proper refactoring
* Contains anti-recursion hack in the new SMGR code, which is needed until we no longer store the metadata in the catalog


dAdAbird (approved):

src/access/pg_tde_tdemap.c
```diff
@@ -108,20 +108,23 @@ static void finalize_key_rotation(char *m_path_old, char *k_path_old, char *m_pa
...
 	XLogRelKey xlrec;
 
-	master_key = GetMasterKey();
+	pg_tde_set_db_file_paths(newrlocator->dbOid);
+	master_key = GetMasterKey(newrlocator->dbOid);
 	if (master_key == NULL)
 	{
 		ereport(ERROR,
-				(errmsg("failed to retrieve master key")));
+				(errmsg("1 - failed to retrieve master key")));
```
codeforall, line 125: I believe this 1 is left out as part of some debugging code?

codeforall (approved):
There are a few sinlge line comments that could be changed to C-style ones. Other than that all looks good.

PR #204, opened 2024-06-06
bugfix: use raw fd to lseek

rainhard:
map_file is vitual file descriptor, lseek should use raw fd


dAdAbird (approved):

codeforall (approved):
LGTM

ImTheKai:
Thanks a lot @rainhard for your contribution. Much appreciated. If there is any Swag like a shirt or something, we will happily send it your way.

PR #214, opened 2024-06-14
Add key management for WAL

dAdAbird:
* Make the *.map *.dat processing code aware of custom databases and
  table spaces

* Add XLog GUC and init the keyring based on that. Only FS for now

* Make the internal/external key infrastructure work with custom
  (not stored in the database) keyrings.

* Check and create an internal key for XLog during the server start.
If the key is created (not the first start with the EncryptWAL), then
upload it into the cache. We can't read the key from files while
writing the XLog to the disk as it happens in the critical section and
no palloc is allowed.

* Create a custom cache for the global catalog external key as we can't
use PG's hashmap during the (again, no pallocs in critical section).

* During the server start, when pg_tde module is loading and it needs to
read *.map, *.dat file, InitFileAccess is yet to be called, hence Vfd
isn't ready to use. The same gonna happen during recovery. So use raw
pread/pwrite calls istead.


src/access/pg_tde_tdemap.c
```diff
@@ -436,33 +444,42 @@ pg_tde_file_header_write(char *tde_filename, File tde_file, TDEMasterKeyInfo *ma
...
-static File
-pg_tde_file_header_read(char *tde_filename, File tde_file, TDEFileHeader *fheader, bool *is_new_file, off_t *bytes_read)
+static int
+pg_tde_file_header_read(char *tde_filename, int fd, TDEFileHeader *fheader, bool *is_new_file, off_t *bytes_read)
 {
 	Assert(fheader);
 
-	*bytes_read = FileRead(tde_file, fheader, TDE_FILE_HEADER_SIZE, 0, WAIT_EVENT_DATA_FILE_READ);
+	/* TODO: pgstat_report_wait_start / pgstat_report_wait_end */
+	*bytes_read = pg_pread(fd, fheader, TDE_FILE_HEADER_SIZE, 0);
```
dutow, line 477: Can we use some dyanmic switching between FileRead/pg_pread and other functions (`if(filestuffinitialilzed) { doone; } else { dotheother; }`)?

src/catalog/tde_global_catalog.c
```diff
@@ -0,0 +1,235 @@
...
+{
+	return GetRelationKeyWithKeyring(GLOBAL_SPACE_RLOCATOR(obj_id), EncryptionState->keyring);
+}
+
+/* TODO: add Vault */
+static void
+init_keyring(void)
+{
+	EncryptionState->keyring->type = get_keyring_provider_from_typename(KRingProviderType);
+	switch (EncryptionState->keyring->type)
```
dutow, line 160: I'm not sure about implementing a separate keyring mechanism for the global catalog (we have to refactor the other anyway), but I definitely don't like duplicating the configuration part. Can't we just extract `load_keyring_provider_options` and reuse it here? (and similarly to files, if the postgres JSON API doesn't work during initialization/recovery, we can't use it) 

We can keep this as-is with a TODO to fix this as part of the keyring metadata recovery, but then it needs those todos (and issues)

src/access/pg_tde_tdemap.c
dAdAbird: Great idea. I'll look into that.

src/catalog/tde_global_catalog.c
dAdAbird: I've added TODO and created an issue https://github.com/Percona-Lab/pg_tde/issues/218

src/access/pg_tde_tdemap.c
dAdAbird: Unfortunately there is no way to detect it w/o core changes. The backend checks it by asserting static variable and there is no external function without side effects that we can use.

src/catalog/tde_global_catalog.c
```diff
@@ -0,0 +1,243 @@
...
+				 errmsg("could not generate internal key for \"WAL\": %s",
+						ERR_error_string(ERR_get_error(), NULL))));
+	}
+
+	rlocator = &GLOBAL_SPACE_RLOCATOR(XLOG_TDE_OID);
+	rel_key_data = tde_create_rel_key(rlocator->relNumber, &int_key, &mkey->keyInfo);
+	enc_rel_key_data = tde_encrypt_rel_key(mkey, rel_key_data, rlocator);
+	pg_tde_write_key_map_entry(rlocator, enc_rel_key_data, &mkey->keyInfo);
+
+	pg_tde_put_key_into_map(rlocator->relNumber, rel_key_data);
```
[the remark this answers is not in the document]
dAdAbird: Good suggestion.
My take on it:
We need an Xlog key to be stored in the top memory context because of the [InternalKey.ctx](https://github.com/Percona-Lab/pg_tde/blob/fefe7b7dead0b19b35146389f10af0936e0e1e4b/src/include/access/pg_tde_tdemap.h#L20). This context if being initialized by OpenSSL with the pointer to the encryption context. And should the cache be in the shared memory, new backends would try to use the context of the first initialized backend (hence segfault). Having the case in the top memory context means every backend inherits it with the NULL context and initializes it exclusively. Using shared memory here would mean a pool of contexts, locking etc.
But we can have a global catalog cache in TopMemory context. I just don't want to do it in the PR, the reasons being:
1. The PR is meant to be a minimum working version and it is already over bloated.
2. I'd like to address it along with the key rotation as it may add some additional challenges and requirements for the keys caching.

Another thing: I'd change InternalKey cache in general. Now it is a linked list with objects allocated all over the memory and traversing such a list is an absolute performance nightmare (CPU caches etc).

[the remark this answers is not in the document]
dAdAbird: I've added a TODO comment

dutow (approved):

codeforall (approved):

PR #230, opened 2024-07-03
Remove 'percona_tde.pg_tde_key_provider' user catalog

codeforall:
Remove 'percona_tde.pg_tde_key_provider' user catalog and introduce provider info file for key providers

This commit removes the 'percona_tde.pg_tde_key_provider' user catalog and replaces it with a provider info file for saving key providers. This change ensures that the key provider information can be accessed during recovery, as the user catalogs cannot be relied upon in such scenarios.

The commit maintains the current API functions, so callers will not experience any differences in functionality or usage after this change.

Additionally, the commit adjusts how the shared memory manager retrieves information about the number of LWLocks required by the extension, optimizing the process.

TODO: Implement xlog message for cleaning up the provider info file during recovery operations to ensure consistency and avoid potential issues.

### DIFF
--- a/pg_tde--1.0.sql
+++ b/pg_tde--1.0.sql
@@ -3,37 +3,16 @@
 -- complain if script is sourced in psql, rather than via CREATE EXTENSION
 \echo Use "CREATE EXTENSION pg_tde" to load this file. \quit
 
--- pg_tde catalog tables
-CREATE SCHEMA percona_tde;
--- Note: The table is created using heap storage becasue we do not want this table
--- to be encrypted by pg_tde. This table is used to store key provider information
--- and we do not want to encrypt this table using pg_tde.
-CREATE TABLE percona_tde.pg_tde_key_provider(provider_id SERIAL,
-        keyring_type VARCHAR(10) CHECK (keyring_type IN ('file', 'vault-v2')),
-        provider_name VARCHAR(255) UNIQUE NOT NULL, options JSON, PRIMARY KEY(provider_id)) using heap;
-
--- If you want to add new provider types, you need to make appropriate changes
--- in include/catalog/tde_keyring.h and src/catalog/tde_keyring.c files.
-
-SELECT pg_catalog.pg_extension_config_dump('percona_tde.pg_tde_key_provider', '');
-
--- Trigger function to check principal key dependency on key provider row
-CREATE FUNCTION keyring_delete_dependency_check_trigger()
-RETURNS TRIGGER
+-- Key Provider Management
+CREATE FUNCTION pg_tde_add_key_provider_internal(provider_type VARCHAR(10), provider_name VARCHAR(128), options JSON)
+RETURNS INT
 AS 'MODULE_PATHNAME'
 LANGUAGE C;
-
-CREATE TRIGGER pg_tde_key_provider_delete_dependency_check_trigger
-BEFORE DELETE ON percona_tde.pg_tde_key_provider
-FOR EACH ROW
-EXECUTE FUNCTION keyring_delete_dependency_check_trigger();
-
--- Key Provider Management
 
 CREATE OR REPLACE FUNCTION pg_tde_add_key_provider(provider_type VARCHAR(10), provider_name VARCHAR(128), options JSON)
 RETURNS INT
 AS $$
-    INSERT INTO percona_tde.pg_tde_key_provider (keyring_type, provider_name, options) VALUES (provider_type, provider_name, options) RETURNING provider_id;
+    SELECT pg_tde_add_key_provider_internal(provider_type, provider_name, options);
 $$
 LANGUAGE SQL;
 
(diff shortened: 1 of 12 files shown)


pg_tde--1.0.sql
```diff
@@ -3,37 +3,16 @@
 -- complain if script is sourced in psql, rather than via CREATE EXTENSION
 \echo Use "CREATE EXTENSION pg_tde" to load this file. \quit
 
--- pg_tde catalog tables
-CREATE SCHEMA percona_tde;
--- Note: The table is created using heap storage becasue we do not want this table
--- to be encrypted by pg_tde. This table is used to store key provider information
--- and we do not want to encrypt this table using pg_tde.
-CREATE TABLE percona_tde.pg_tde_key_provider(provider_id SERIAL,
```
dutow, line 11: Shouldn't we replace this with a view that provides the same functionality?

src/common/pg_tde_utils.c
```diff
@@ -215,3 +216,15 @@ extract_json_option_value(Datum top_json, const char* field_name)
...
+
+/* returns the palloc'd string */
+char *
+pg_tde_get_tde_file_dir(Oid dbOid, Oid spcOid)
+{
+	/* If this is a global space, than the call might be in a critial section
+	 * (during XLog write) so we can't do GetDatabasePath as it calls palloc()
+	 */
+	if (spcOid == GLOBALTABLESPACE_OID)
+		return pstrdup("global");
```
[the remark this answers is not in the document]
codeforall: I just changed the existing function and added pstrdup as we were leaking the memory previously without realising the comments also needed a fix. 
Thanks for the update

pg_tde--1.0.sql
```diff
@@ -3,37 +3,16 @@
...
 LANGUAGE C;
 
-CREATE TRIGGER pg_tde_key_provider_delete_dependency_check_trigger
-BEFORE DELETE ON percona_tde.pg_tde_key_provider
-FOR EACH ROW
-EXECUTE FUNCTION keyring_delete_dependency_check_trigger();
-
--- Key Provider Management
-
 CREATE OR REPLACE FUNCTION pg_tde_add_key_provider(provider_type VARCHAR(10), provider_name VARCHAR(128), options JSON)
```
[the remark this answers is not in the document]
codeforall: I think it's a good idea. We can take it up as a separate PR.

src/catalog/tde_keyring.c
```diff
@@ -49,11 +49,89 @@ PG_FUNCTION_INFO_V1(keyring_delete_dependency_check_trigger);
...
+static inline LWLock *
+tde_provider_info_lock(void)
+{
+	Assert(sharedPrincipalKeyState);
+	return &sharedPrincipalKeyState->Locks[TDE_LWLOCK_PI_FILES];
+}
+
+void InitializeKeyProviderInfo(void)
+{
+	ereport(LOG, (errmsg("Initializing TDE principal key info")));
```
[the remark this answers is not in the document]
codeforall: Fixed for all messages added as part of this PR!
We might also need to do a code run-through and fix similar inconsistencies in other parts of the code.

pg_tde--1.0.sql
codeforall: I believe we would need to provide a view and edit function as mentioned in the comment https://github.com/Percona-Lab/pg_tde/pull/230#discussion_r1665792565, but we can handle it as a separate PR

src/catalog/tde_keyring.c
```diff
@@ -107,103 +167,39 @@ load_keyring_provider_from_tuple(HeapTuple tuple, TupleDesc tupDesc)
...
-		keyring = load_keyring_provider_from_tuple(tuple, tupDesc);
-		break; /* We expect only one record */
+		keyring = (GenericKeyring *)linitial(providers);
+		list_free(providers);
+	}
+	else
+	{
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				 errmsg("Key provider \"%s\" does not exists", provider_name),
```
dAdAbird, line 187:
```suggestion
				 errmsg("key provider \"%s\" does not exists", provider_name),
```

src/catalog/tde_keyring.c
```diff
@@ -267,99 +265,225 @@ load_vaultV2_keyring_provider_options(Datum keyring_options)
...
+	fd = BasicOpenFile(kp_info_path, PG_BINARY);
+	if (fd < 0)
+	{
+		LWLockRelease(tde_provider_info_lock());
+		ereport(DEBUG2,
+			(errcode_for_file_access(),
+				errmsg("could not open tde file \"%s\": %m", kp_info_path)));
+		return NIL;
+	}
+	/* we also need to verify the name conflixt and generate the next provider ID */
```
dAdAbird, line 417: I don't think this comment is relevant here
```suggestion
```

PR #238, opened 2024-07-10
WAL keyring improvements

dAdAbird:
- Rename database key rotation functions to make room for the global space ones.
- Now, during the first start, we would create a default temporary key provider for the global space. A user can (and should) create their own key provider afterwards. This allows use the same codepath and internal interfaces for the keyring management across databases and the global space.
- Now need to cache the principal key for the global space as we use it only at the server start to decrypt internal key. Then internal key persists in the memory cache.

Fixes https://perconadev.atlassian.net/browse/PG-835, https://perconadev.atlassian.net/browse/PG-833

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ CREATE EXTENSION pg_tde;
    
 ```sql
 -- For Vault-V2 key provider
-pg_tde_add_key_provider_vault_v2(
+pg_tde_add_database_key_provider_vault_v2(
                         provider_name VARCHAR(128),
                         vault_token TEXT,
                         vault_url TEXT,
@@ -55,25 +55,25 @@ pg_tde_add_key_provider_vault_v2(
                         vault_ca_path TEXT);
 
 -- For File key provider
-FUNCTION pg_tde_add_key_provider_file(
+FUNCTION pg_tde_add_database_key_provider_file(
                         provider_name VARCHAR(128), 
                         file_path TEXT);
 
 ```
 **Example**: Add a `File` key provider and name it `file`.
 ```sql
-SELECT pg_tde_add_key_provider_file('file','/tmp/pgkeyring');
+SELECT pg_tde_add_database_key_provider_file('file','/tmp/pgkeyring');
 ```
 **Note: The `File` provided is intended for development and stores the keys unencrypted in the specified data file.**
-6. Set the principal key for the database using the `pg_tde_set_database_key` function.
+6. Set the principal key for the database using the `pg_tde_set_database_principal_key` function.
 ```sql
-FUNCTION pg_tde_set_database_key (
+FUNCTION pg_tde_set_database_principal_key (
                 principal_key_name VARCHAR(255), 
                 provider_name VARCHAR(255));
 ```
 **Example**: Set the principal key named `my-principal-key` using the `file` as a key provider.
 ```sql
-SELECT pg_tde_set_database_key('my-principal-key','file');
+SELECT pg_tde_set_database_principal_key('my-principal-key','file');
 ```
 
 7. You are all set to create encrypted tables. For that, specify `USING pg_tde_basic` access method in the `CREATE TABLE` statement.
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ CREATE EXTENSION pg_tde;
    
 ```sql
 -- For Vault-V2 key provider
-pg_tde_add_database_key_provider_vault_v2(
+pg_tde_add_key_provider_vault_v2(
                         provider_name VARCHAR(128),
                         vault_token TEXT,
                         vault_url TEXT,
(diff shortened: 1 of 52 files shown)


README.md
```diff
@@ -65,15 +65,15 @@ FUNCTION pg_tde_add_key_provider_file(
 SELECT pg_tde_add_key_provider_file('file','/tmp/pgkeyring');
 ```
 **Note: The `File` provided is intended for development and stores the keys unencrypted in the specified data file.**
-6. Set the principal key for the database using the `pg_tde_set_principal_key` function.
+6. Set the principal key for the database using the `pg_tde_set_database_key` function.
```
dutow, line 68: Now actually looking at this change, I don't think this is a good idea. We just agreed to name "master key" to "principal key" in the beta, and then we would rename all public occurrences again to "database key". Documentation / internal code would still say principal key, but the functions that actually deal with the keys would all be using the database key name.

dAdAbird: Yeah, I had the same thoughts tbh. Should we have `pg_tde_set_principal_db_key`/ `pg_tde_set_principal_global_space_key` etc. instead?
The renaming is a one part and the other is creating respective functions for global spaces. But then something like `pg_tde_set_principal_global_space_key` becomes ridiculously long. Having just `...principal_global_key` (w/o space) looks a bit confusing. And I'm starting to consider having `pg_tde_set_principal_key()` with the enum option `DATABASE`, `GLOBAL_SPACE`. But I'm not sure.

dAdAbird: On another note, not related to this commit, but it might be a right time to do something about that:
Looking at `pg_tde--1.0.sql` we have inconsistency with the functions naming there. Most time they are prefixed with `pg_tde_` but there are `pg_tdeam_handler`, `pg_td2eam_handler`, and `pgtde_is_encrypted` which don't follow that convention

documentation/docs/functions.md
```diff
@@ -33,43 +33,43 @@ where:
 
 All parameters can be either strings, or JSON objects [referencing remote parameters](external-parameters.md).
 
-## pg_tde_set_principal_key
+## pg_tde_set_database_key
```
dutow, line 36: database_principal_key?

documentation/docs/functions.md
```diff
@@ -33,43 +33,43 @@ where:
...
 Sets the principal key for the database using the specified key provider.
 
 The principal key name is also used for constructing the name in the provider, for example on the remote Vault server.
 
-You can use this function only to a principal key. For changes in the principal key, use the [`pg_tde_rotate_key`](#pg_tde_rotate_key) function.
+You can use this function only to a principal key. For changes in the principal key, use the [`pg_tde_rotate_database_key`](#pg_tde_rotate_database_key) function.
 
 ```sql
-SELECT pg_tde_set_principal_key('name-of-the-principal-key', 'provider-name');
+SELECT pg_tde_set_database_key('name-of-the-principal-key', 'provider-name');
```
dutow, line 45: and here

src/catalog/tde_principal_key.c
```diff
@@ -735,12 +742,67 @@ pg_tde_rotate_key(PG_FUNCTION_ARGS)
...
+    TDEPrincipalKey *current_key;
+
+    if (!PG_ARGISNULL(0))
+        new_principal_key_name = text_to_cstring(PG_GETARG_TEXT_PP(0));
+    if (!PG_ARGISNULL(1))
+        new_provider_name = text_to_cstring(PG_GETARG_TEXT_PP(1));
+    ensure_new_key = PG_GETARG_BOOL(2);
+
+
+    ereport(LOG, (errmsg("Rotating principal key to [%s : %s] for the database", new_principal_key_name, new_provider_name)));
```
dutow, line 768: for the cluster, or maybe explicitly mention global key?

src/catalog/tde_keyring.c
```diff
@@ -499,7 +511,23 @@ get_keyring_infofile_path(char* resPath, Oid dbOid, Oid spcOid)
 }
 
 Datum
-pg_tde_add_key_provider_internal(PG_FUNCTION_ARGS)
+pg_tde_add_database_key_provider_internal(PG_FUNCTION_ARGS)
```
dutow, line 514: We could avoid all this code duplication with a simple helper function, only 1 line is different in the two variants.

src/catalog/tde_keyring.c
```diff
@@ -504,12 +514,23 @@ pg_tde_add_key_provider_internal(PG_FUNCTION_ARGS)
 	char *provider_type = text_to_cstring(PG_GETARG_TEXT_PP(0));
 	char *provider_name = text_to_cstring(PG_GETARG_TEXT_PP(1));
 	char *options = text_to_cstring(PG_GETARG_TEXT_PP(2));
+	bool is_global = PG_GETARG_BOOL(3);
 	KeyringProvideRecord provider;
+	Oid dbOid = MyDatabaseId;
+	Oid spcOid = MyDatabaseTableSpace;
+
+#ifdef PERCONA_FORK
```
dutow, line 522: Do we really need this ifdef here?

src/catalog/tde_principal_key.c
```diff
@@ -718,29 +724,63 @@ Datum pg_tde_set_principal_key(PG_FUNCTION_ARGS)
...
+    Oid spcOid = MyDatabaseTableSpace;
 
     if (!PG_ARGISNULL(0))
         new_principal_key_name = text_to_cstring(PG_GETARG_TEXT_PP(0));
     if (!PG_ARGISNULL(1))
         new_provider_name = text_to_cstring(PG_GETARG_TEXT_PP(1));
     ensure_new_key = PG_GETARG_BOOL(2);
+    is_global = PG_GETARG_BOOL(3);
 
+#ifdef PERCONA_FORK
```
dutow, line 747: Same question, do we need this ifdef?

src/catalog/tde_keyring.c
dAdAbird: Yep, you're right, we can get rid of it: https://github.com/Percona-Lab/pg_tde/pull/238/commits/cf02b52dcd232e99dd46c8d11ec4b7e8522d4e74

dutow (approved):
Looks good, please make sure to use squash and merge, so we don't get this long commit history into the main branch.

PR #239, opened 2024-07-12
Add function to list available key provides for database

codeforall:


pg_tde--1.0.sql
```diff
@@ -56,6 +56,15 @@ AS $$
 $$
 LANGUAGE SQL;
 
+CREATE FUNCTION pg_tde_list_all_key_providers
```
dutow, line 59: Generally the function looks good, but there's one more thing I'm thinking about: since this returns possibly sensitive data, shouldn't we limit who can execute the function? And this is also a question for our other functions.

codeforall: This is a valid point. The same problem exists in all pg_tde functions, I've created an issue for that.
https://perconadev.atlassian.net/browse/PG-853

src/catalog/tde_keyring.c
```diff
@@ -513,3 +532,59 @@ pg_tde_add_key_provider_internal(PG_FUNCTION_ARGS)
...
+	Tuplestorestate *tupstore;
+	TupleDesc tupdesc;
+	MemoryContext per_query_ctx;
+	MemoryContext oldcontext;
+	ReturnSetInfo *rsinfo = (ReturnSetInfo *)fcinfo->resultinfo;
+	/* check to see if caller supports us returning a tuplestore */
+	if (rsinfo == NULL || !IsA(rsinfo, ReturnSetInfo))
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("pg_tde_list_all_key_providers: Set-valued function called in context that cannot accept a set.")));
```
dAdAbird, line 550: Here and below, looking through the PG code I have two style notices regarding error messages:
1. The only time there is a capital letter after the colon `:`, is when it is an acronym with all caps.
2. The error message never finishes with the full stop `.`

So the message should be rather:
`"pg_tde_list_all_key_providers: set-valued function called in context that cannot accept a set"`

codeforall: Fixed!

dAdAbird (approved):

PR #245, opened 2024-07-23
Change access method to pg_tde_basic

ImTheKai:
Now that we do have the pg_tde and the pg_tde_basic access method, we need to change the default one to pg_tde_basic in the documentation, to avoid running into deployment issues.

Once our patched version is ready, we'll add more details around the pg_tde access method.


nastena1606 (approved):
LGTM

dutow (approved):

PR #252, opened 2024-08-07
PG-857: use common JSON API for keyring

dAdAbird:
This commit replaces dependency of the keyring code on JSON backend functions with common JSON API.

Usage of the backend JSON funcs prevents the code to be used by frontend tools like pg_waldump. Besides it requiers extra conversions to Datums and DirectFunctionCall-s.

For: https://perconadev.atlassian.net/browse/PG-857

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -44,6 +44,7 @@ src/keyring/keyring_vault.o \
 src/keyring/keyring_api.o \
 src/catalog/tde_global_space.o \
 src/catalog/tde_keyring.o \
+src/catalog/tde_keyring_parse_opts.o \
 src/catalog/tde_principal_key.o \
 src/common/pg_tde_shmem.o \
 src/common/pg_tde_utils.o \
--- a/meson.build
+++ b/meson.build
@@ -43,6 +43,7 @@ pg_tde_sources = files(
 
         'src/catalog/tde_global_space.c',
         'src/catalog/tde_keyring.c',
+        'src/catalog/tde_keyring_parse_opts.c',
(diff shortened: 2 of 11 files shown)


t/004_file_config.pl
```diff
@@ -23,10 +23,6 @@
 print $conf "shared_preload_libraries = 'pg_tde'\n";
 close $conf;
 
-open my $conf2, '>>', "/tmp/datafile-location";
```
dutow, line 26: Why? I think the issue here is that "file" remote support is still missing from the code, plese read the documentation: https://percona-lab.github.io/pg_tde/main/external-parameters.html

dutow (commented):
Looks working, and okay, but...

My main issue with this pull request is that now instead of a nice, simple and easy to understand JSON code we end up with a long and complex thing that  works, but it is difficult to understand and/or modify.

I already had this issue with the previous JSON refactoring where we removed jsonc, because that already made the code more complex, but now it is even more so.

t/004_file_config.pl
dAdAbird: Remote is supported but with URL only and not path. It didn't make sense to me at first and I thought this is some leftovers (was about to clarify in it). 
I'll do the "remote path" but the question is should the 'type' parameter be like in tests `file` or like in docs `remote`?

test L33:
```
json_object( 'type' VALUE 'file', 'path' VALUE '/tmp/datafile-location' )
```
docs:
```
json_object( 'type' VALUE 'remote', 'path' VALUE '/tmp/datafile-location' )
```

dutow (approved):

PR #255, opened 2024-08-07
Adding a workflow for PPG Server installation & tap test case for testing pg_tde along with install of multiple PPG extensions.

Naeem-Akhter:
Adding a workflow for PPG Server installation & tap test case for testing pg_tde along with install of multiple PPG extensions.

### DIFF
--- a/t/005_multiple_extensions.pl
+++ b/t/005_multiple_extensions.pl
@@ -1,0 +1,158 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use File::Basename;
+use File::Compare;
+use File::Copy;
+use Test::More;
+use lib 't';
+use pgtde;
+
+# Get file name and CREATE out file name and dirs WHERE requried
+PGTDE::setup_files_dir(basename($0));
+
+my $PG_VERSION_STRING = `pg_config --version`;
+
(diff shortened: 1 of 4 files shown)


t/005_multiple_extensions.pl
```diff
@@ -0,0 +1,158 @@
...
+# DROP EXTENSION
+$stdout = $node->safe_psql('postgres', 'DROP EXTENSION pg_stat_monitor;', extra_params => ['-a']);
+ok($cmdret == 0, "DROP PGTDE EXTENSION");
+PGTDE::append_to_debug_file($stdout);
+
+# Stop the server
+$node->stop();
+
+# compare the expected and out file
+#my $compare = PGTDE->compare_results();
```
dutow, line 152: Generally looks good , but why is this comparison commented out? Also, the expected file is missing from the PR.

Naeem-Akhter: Intentions was only look for server crash in this test case with PPG distro extensions present along with pg_tde, rest of logic is same as 004 tap test case. Nothing wrong in enabling the comparison, I will do so.

Naeem-Akhter: @dutow done.

dutow (approved):

PR #263, opened 2024-08-20
Fix smgr for PG17

dAdAbird:
PG_17_STABLE added new method to "f_sgmr" interface. That method was lacking in our tde_smgr which led to SIGSEGV when core tried to use it.


src16/smgr/pg_tde_smgr.c
```diff
@@ -0,0 +1,218 @@
+
```
dAdAbird, line 1: @dutow, @codeforall I'm not sure if need this back port for PG16. Do we?

dAdAbird: we probably do need it at least for tests

dutow: No need for it, the smgr atch is PG17 only, this shouldn't be in the src16 folder.

dutow (approved):

PR #265, opened 2024-08-26
Performance / memory usage fix: do not allocate memory in pg_tde_slot

dutow:
Until now, for some reason we allocated memory for each decrypted tuple in pg_tde_slot, and only freed all that memory when the transaction ended. This caused two issues:

* palloc takes time, and the many small pallocs resulted in a significant performance drop in scans
* sequential scans could potentially allocate memory to the entire table, requiring way too much memory for large tables

This allocation is most likely a leftover from before we even used slots, and handled selects differently. With slots, we don't need them at all, as slots are not expected to handle multiple tuples at the same time, only the last one.

This commit removes the palloc completely, and instead adds a single BLCKSZ array to the slot structure, which can hold any size of decrypted tuple.

To be safe, it also disables the get tuple function, forcing the core code to use copy instead when needed.

This change results in:

1. no "memory spike" during sequential scans
2. ~1.55x overhead instead of ~2.2x

(The TODO in slot_copytuple will be addressed in a separate commit)

### DIFF
--- a/src/access/pg_tde_slot.c
+++ b/src/access/pg_tde_slot.c
@@ -54,6 +54,8 @@ tdeheap_tts_buffer_heap_init(TupleTableSlot *slot)
 static void
 tdeheap_tts_buffer_heap_release(TupleTableSlot *slot)
 {
+	TDEBufferHeapTupleTableSlot *bslot = (TDEBufferHeapTupleTableSlot *) slot;
+	bslot->decrypted_tuple = NULL;
 }
 
 static void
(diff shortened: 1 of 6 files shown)


dAdAbird (approved):

codeforall:
Overall, the PR looks good. Just one minor comment. I believe we can now get rid of TdeSlotForgetDecryptedTuple function as we do not need to pfree the decrypted_tuple

codeforall (approved):

PR #267, opened 2024-08-26
Fix tap test

dAdAbird:
This test was created before but merged after the access method rename...


dutow (approved):

dutow:
Approved, but looks like the out file also needs updating

PR #277, opened 2024-09-05
PG-853: Access control of pg_tde SQL functions

codeforall:
Add SQL interfaces for granting and revoking access to key management and viewer functions. This commit introduces four new SQL functions to manage access to key-related functionalities in the `pg_tde` extension:

- `tde_grant_key_management_to_role`: Grants execute permissions on key management functions to the specified user or role.
- `tde_revoke_key_management_from_role`: Revokes execute permissions on key management functions from the specified user or role.
- `tde_grant_key_viewer_to_role`: Grants execute permissions on key viewer functions to the specified user or role.
- `tde_revoke_key_viewer_from_role`: Revokes execute permissions on key viewer functions from the specified user or role.

Additionally, upon creating the extension, all execute permissions are revoked from the `PUBLIC` role. Therefore, a superuser must explicitly grant the necessary permissions to non-superusers to access these functions after the extension is created.

These additions provide a more controlled and secure way to manage permissions for key management and viewer functionalities within the extension.

### DIFF
--- a/pg_tde--1.0.sql
+++ b/pg_tde--1.0.sql
@@ -275,7 +275,7 @@ $$;
 SELECT pg_tde_extension_initialize();
 
 
-CREATE OR REPLACE FUNCTION tde_grant_execute_privilege_on_function(
+CREATE OR REPLACE FUNCTION pg_tde_grant_execute_privilege_on_function(
     target_user_or_role TEXT,
     target_function_name TEXT,
     target_function_args TEXT
@@ -296,7 +296,7 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION tde_revoke_execute_privilege_on_function(
+CREATE OR REPLACE FUNCTION pg_tde_revoke_execute_privilege_on_function(
     target_user_or_role TEXT,
     target_function_name TEXT,
     argument_types TEXT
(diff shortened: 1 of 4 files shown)


pg_tde--1.0.sql
```diff
@@ -274,3 +274,178 @@ $$;
...
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_rotate_principal_key', 'pg_tde_global, varchar, varchar');
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_rotate_principal_key', 'varchar, varchar');
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_rotate_principal_key_internal', 'varchar, varchar, BOOLEAN, BOOLEAN');
+
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_list_all_key_providers', 'OUT INT, OUT varchar, OUT varchar, OUT JSON');
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_is_encrypted', 'VARCHAR');
+
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_principal_key_info_internal', 'BOOLEAN');
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_principal_key_info', '');
+    PERFORM tde_grant_execute_privilege_on_function(target_user_or_role, 'pg_tde_principal_key_info', 'pg_tde_global');
```
dAdAbird, line 352: Should we use just
```sql
PERFORM tde_grant_key_viewer_to_role(target_user_or_role);
```
instead? It'll be easier to read and maintain both functions as for me

dAdAbird:
Another question:
Can we just grant/revoke access to only C functions like `*_internal` etc? So makes less `PERFORM tde_grant_execute_privilege_on_function ...` code. Hence easier to follow and less chances to make mistake maintaining `tde_grant_*`, `tde_revoke_*` functions.

I mean, shouldn't revoke access to `pg_tde_add_key_provider_internal` implicitly make `pg_tde_add_key_provider_file` and `pg_tde_add_key_provider_vault_v2` functions useless?

dutow:
Why `tde_*` instead of `pg_tde_*`, which we used for the other functions so far? Shouldn't these use the same prefix?

t/007_access_control.pl
```diff
@@ -0,0 +1,104 @@
...
+ok($rt_value == 1, "Restart Server");
+
+# TRY performing operations without permission
+PGTDE::append_to_file("-- pg_tde_add_key_provider_file should throw access denied");
+($cmdret, $stdout, $stderr) = $node->psql('postgres', "SELECT pg_tde_add_key_provider_file('file-vault','/tmp/pg_tde_test_keyring.per');", extra_params => ['-a', '-U', 'test_access']);
+PGTDE::append_to_file($stderr);
+
+PGTDE::append_to_file("-- pg_tde_set_principal_key should also fail");
+($cmdret, $stdout, $stderr) = $node->psql('postgres', "SELECT pg_tde_set_principal_key('test-db-principal-key','file-vault');", extra_params => ['-a', '-U', 'test_access']);
+PGTDE::append_to_file($stderr);
```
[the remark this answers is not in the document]
codeforall: Done!

pg_tde--1.0.sql
codeforall: Done!

codeforall:
> Why `tde_*` instead of `pg_tde_*`, which we used for the other functions so far? Shouldn't these use the same prefix?



> Another question: Can we just grant/revoke access to only C functions like `*_internal` etc? So makes less `PERFORM tde_grant_execute_privilege_on_function ...` code. Hence easier to follow and less chances to make mistake maintaining `tde_grant_*`, `tde_revoke_*` functions.
> 
> I mean, shouldn't revoke access to `pg_tde_add_key_provider_internal` implicitly make `pg_tde_add_key_provider_file` and `pg_tde_add_key_provider_vault_v2` functions useless?

The reason I'm performing the grant/revoke on every function is to ensure that if a "permission denied" error occurs, it reflects the name of the function the user actually called. Simply revoking access from the *_internal function would also trigger a permission denied error, but it would always refer to the internal function that the user did not directly execute.

dAdAbird (approved):

dutow (approved):

PR #282, opened 2024-09-17
PMM-1035 Fix toast slice fetching with offset

artemgavrilov:
https://perconadev.atlassian.net/browse/PG-1035

### DIFF
--- a/expected/toast_extended_storage.out
+++ b/expected/toast_extended_storage.out
@@ -77,4 +77,27 @@ UPDATE indtoasttest SET cnt = cnt +1, f1 = f1||'' RETURNING substring(indtoastte
 
 UPDATE indtoasttest SET  f2 = '+'||f2||'-' ;
 DROP TABLE indtoasttest;
+-- Test substr with toasted externalized bytea values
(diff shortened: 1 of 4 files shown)


dAdAbird (approved):
Yes, `sliceoffset` is redundant in the decryption offset. `sliceoffset` points to the offset from the start of TOAST value and `(curchunk * TOAST_MAX_CHUNK_SIZE - sliceoffset) + chcpystrt` will get the offset for each chunk from the start of fetched data. But for decryption, we need the offset relative to the beginning of TOAST data.
For example:
`SELECT substring(t, 6660, 5984) FROM toasttest;` will get 4 chunks and `sliceoffset` would be 6659. And offset with the `sliceoffset` correlation for these chunks would be 0, 1325, 3321, and 5317 - offset relative to the beginning of the requested data - which is what we need for memcpy data into the buffer. But for decryption, we need 6659, 7984, 9980, and 11976 offsets - relative to the beginning of the TOASTed data.

artemgavrilov:
@dAdAbird Thanks for an explanation!

dutow (approved):

PR #287, opened 2024-09-18
PG-952: Handle non existent principal key in tde_heap

dutow:
Previously when the database had no principal key, tde_heap simply igored it and created a table without encryption.

With this change, it reports an error instead.

### DIFF
--- a/src/pg_tde_event_capture.c
+++ b/src/pg_tde_event_capture.c
@@ -21,6 +21,9 @@
 #include "commands/event_trigger.h"
 #include "common/pg_tde_utils.h"
 #include "pg_tde_event_capture.h"
+#include "commands/tablespace.h"
+#include "catalog/tde_principal_key.h"
+#include "miscadmin.h"
 
 /* Global variable that gets set at ddl start and cleard out at ddl end*/
 TdeCreateEvent tdeCurrentCreateEvent = {.relation = NULL};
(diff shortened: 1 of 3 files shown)


src/pg_tde_event_capture.c
```diff
@@ -105,6 +110,17 @@ pg_tde_ddl_command_start_capture(PG_FUNCTION_ARGS)
...
 			tdeCurrentCreateEvent.encryptMode = true;
 		}
+
+		tablespace_oid = stmt->tablespacename != NULL ? get_tablespace_oid(stmt->tablespacename, false) 
+						 : MyDatabaseTableSpace;  
+		principal_key = GetPrincipalKey(MyDatabaseId, tablespace_oid);
+		if (principal_key == NULL)
+		{
+			ereport(ERROR,
+					(errmsg("failed to retrieve principal key")));
```
codeforall, line 120: Overall, it looks good. However, the error message should be more descriptive and suggestive. Maybe an error hint should be added as well to let users know how they can solve this issue, like suggesting they create a principal key.

codeforall (changes requested):

src/pg_tde_event_capture.c
dutow: maybe, I just copied the error message we display for tde_heap_basic

dAdAbird (approved):

PR #292, opened 2024-09-23
PG-1056 xmin and xmax correctness

artemgavrilov:
https://perconadev.atlassian.net/browse/PG-1056


dAdAbird:
@dutow see my commit's message.
1. [Here](https://github.com/Percona-Lab/pg_tde/commit/058666090e0c31d676049d020a8a10df94ed2fe0) it says:
_"To be safe, it also disables the get tuple function, forcing the core code to use copy instead when needed."_
But I don't see why would we need this. If get tuple function is NULL it calls backend's `heap_copytuple()` (not our `slot_copytuple()`) which results in extra palloc instead of just passing the pointer. Besides, see my comments.

Am I missing something? 

2. As far as I can tell my commit should not compromise optimizations of https://github.com/Percona-Lab/pg_tde/pull/265. What do you think?

dutow:
@dAdAbird my concern was that if we only get the tuple and do not copy it, we might find issues because we reuse the same memory again for the next tuple. If I understand the API correctly, this shouldn't happen, but it seemed safer to disable get, and force copy. Looks like I was wrong about that.

artemgavrilov:
Before merge I think we should remove test it this PR because it's isolated case from PG regression suite. Not sure that we need it in tde repo. What do you think?

dutow (approved):

codeforall (approved):

PR #293, opened 2024-09-24
PG-1058 Mergejoin failure

artemgavrilov:
https://perconadev.atlassian.net/browse/PG-1058

```
ERROR:  mergejoin input data is out of order
```


dAdAbird:
Fix in https://github.com/percona/pg_tde/pull/323

PR #307, opened 2024-10-09
PG-1072: Fix major event trigger issues

dutow:
* Create table always checked the principal key and tried to create it, event when we didn't try to create a tde_heap table
* Alter table wasn't handled, and because of this changing a table to tde_heap access method didn't result in an encrypted table.
* defaut_table_access_method wasn't handled, and because of this, creating a table using that also resulted in a non encrypted table.

### DIFF
--- a/src/pg_tde_event_capture.c
+++ b/src/pg_tde_event_capture.c
@@ -24,6 +24,7 @@
 #include "commands/tablespace.h"
 #include "catalog/tde_principal_key.h"
 #include "miscadmin.h"
+#include "access/tableam.h"
 
 /* Global variable that gets set at ddl start and cleard out at ddl end*/
 TdeCreateEvent tdeCurrentCreateEvent = {.relation = NULL};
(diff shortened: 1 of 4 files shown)


src/pg_tde_event_capture.c
```diff
@@ -106,24 +107,69 @@ pg_tde_ddl_command_start_capture(PG_FUNCTION_ARGS)
...
+						(errmsg("failed to retrieve principal key. Create one using pg_tde_set_principal_key before using encrypted tables.")));
 
+			}
 		}
 
 	}
+	else if (IsA(parsetree, AlterTableStmt))
+	{
+		fprintf(stderr, "ALTEEEER\n");
+		fprintf(stderr, "ALTEEEER\n");
```
dAdAbird, line 137: debug leftovers?

src/pg_tde_event_capture.c
```diff
@@ -106,24 +107,69 @@ pg_tde_ddl_command_start_capture(PG_FUNCTION_ARGS)
...
+		Relation	rel = table_open(relationId, lockmode);
+		Oid tablespace_oid = rel->rd_locator.spcOid;
+		ListCell   *lcmd;
+		table_close(rel, lockmode);
+
+		foreach(lcmd, stmt->cmds)
+        	{
+			AlterTableCmd *cmd = (AlterTableCmd *) lfirst(lcmd);
+			if(cmd->subtype == AT_SetAccessMethod && strcmp(cmd->name, "tde_heap")==0) {
+		fprintf(stderr, "TO ENCRYPTED\n");
```
dAdAbird, line 151: debug leftovers?

dAdAbird (approved):
LGTM overall just a couple of printf leftovers

src/pg_tde_event_capture.c
dutow: yes, fixed.

PR #308, opened 2024-10-09
PG-1095 Add format validation CI workflow and format sources

artemgavrilov:
PG-1095

This PR adds code format check job. Also it added Makefile target for code formatting according to PG style (pgindent)


.github/workflows/check.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Configure pg_tde
+        run: ./configure
+        working-directory: src/pg_tde
+
+      - name: Checkout cppcheck sources
+        uses: actions/checkout@v4
+        with:
+          repository: "danmar/cppcheck"
+          ref: "2.13.4"
```
artemgavrilov, line 30:
```suggestion
          ref: "2.15"
```

.github/workflows/check.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Configure pg_tde
+        run: ./configure
+        working-directory: src/pg_tde
+
+      - name: Checkout cppcheck sources
+        uses: actions/checkout@v4
+        with:
+          repository: "danmar/cppcheck"
+          ref: "2.15"
```
artemgavrilov, line 30:
```suggestion
          ref: "2.15.0"
```

.github/workflows/check.yaml
```diff
@@ -0,0 +1,113 @@
...
+          cd build
+          cmake ..
+          cmake --build .
+          sudo cmake --install .
+
+      - name: Execute linter check with cppcheck
+        working-directory: src/pg_tde
+        run: |
+          set -x
+          cppcheck --enable=all --inline-suppr --template='{file}:{line},{severity},{id},{message}' --error-exitcode=1 --suppress=missingIncludeSystem --suppress=missingInclude --suppress=unmatchedSuppression:pg_tde.c --check-config .
```
artemgavrilov, line 46:
```suggestion
          cppcheck --enable=all --inline-suppr --template='{file}:{line},{severity},{id},{message}' --error-exitcode=1 --suppress=missingIncludeSystem --suppress=missingInclude --suppress=unmatchedSuppression:pg_tde.c .
```

dutow (changes requested):
Generally looks good, but please exclude the versioned folders (`src16`, `src17`, and later any `srcXX`) from it.

The idea of them is that they only do minimal changes to the files copied from the server, to ensure that we can do minor/major version updates in an automated way using scripts.

Reformatting them would break this workflow.

If some lines/sections are completely added by us, reformatting them is okay, but this should be a manual process, not automated, because of the majority of the changes are done by sed. (the script doing this is in the `tools` directory)

artemgavrilov:
@dutow Done, `srcXX` folders are excluded from formatter

dAdAbird (approved):

dutow (approved):

PR #309, opened 2024-10-10
Refactoring

artemgavrilov:
`pgindent` can't properly handle [compound literals](https://gcc.gnu.org/onlinedocs/gcc-4.3.2/gcc/Compound-Literals.html). It seems that we can easily refactor our code to make code fromatter work.


dAdAbird (approved):

artemgavrilov:
@dutow @codeforall I merged it to unblock my next PR. Please let me know if you have any concerns.

PR #312, opened 2024-10-10
Add SECURITY.md

artemgavrilov:
This security policy the same as we have from PGSM.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,24 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,24 @@
+# Security Policy
+
+## Supported Versions
+
+pg_tde project follows rolling release strategy. So all security updates go to new versions.
```
artemgavrilov, line 5: Is it correct?

ImTheKai: Yes

ImTheKai (approved):

PR #316, opened 2024-10-16
Enhance pg_tde SMGR handling:

codeforall:
Store the encryption key for tde_heap relations in the SmgrRelation to reduce frequent key lookups. Additionally, avoid encrypting relation forks other than MAIN and INIT forks.

### DIFF
--- a/src/smgr/pg_tde_smgr.c
+++ b/src/smgr/pg_tde_smgr.c
@@ -10,6 +10,23 @@
 
 #ifdef PERCONA_EXT
 
+typedef struct TDESMgrRelationData
+{
+	/* parent data */
+	SMgrRelationData reln;
+	/*
+	 * for md.c; per-fork arrays of the number of open segments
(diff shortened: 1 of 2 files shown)


src/smgr/pg_tde_smgr.c
```diff
@@ -62,22 +79,21 @@ static void
...
 
-	if(rkd == NULL)
+	if (!tdereln->encrypted_relation)
 	{
 		mdwritev(reln, forknum, blocknum, buffers, nblocks, skipFsync);
-
-		return;
 	}
-	else
+	else if (forknum == MAIN_FORKNUM || forknum == INIT_FORKNUM)
```
dAdAbird, line 89: Adding this clause, should we have at least and Assert or maybe the final else with ERROR/PANIC to avoid a silent skip in case of something went wrong?

codeforall: Fixed!

PR #317, opened 2024-10-17
Add key type to internal relation keys in pg_tde:

codeforall:
This commit introduces a key type to internal relation keys, allowing segregation of SMGR and heap_basic internal keys. It resolves the issue of double encryption when both tde_heap and tde_heap_basic access methods are in use.
With this enhancement, pg_tde now supports three types of internal keys: SMGR keys, heap_basic keys, and global keys, improving flexibility and preventing redundant encryption operations.

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -60,8 +60,6 @@
 
 #define PG_TDE_FILEMAGIC				0x01454454	/* version ID value = TDE 01 */
 
-#define MAP_ENTRY_FREE					0x00
-#define MAP_ENTRY_VALID					0x01
 
 #define MAP_ENTRY_SIZE					sizeof(TDEMapEntry)
 #define TDE_FILE_HEADER_SIZE			sizeof(TDEFileHeader)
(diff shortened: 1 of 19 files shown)


src/access/pg_tde_tdemap.c
```diff
@@ -680,9 +697,9 @@ pg_tde_perform_rotate_key(TDEPrincipalKey *principal_key, TDEPrincipalKey *new_p
 			continue;
 
 		/* Set the relNumber of rlocator. Ignore the tablespace Oid since we only place our files under the default. */
-		rloc.relNumber = map_entry.relNumber;
+		rloc.relNumber = read_map_entry.relNumber;
 		rloc.dbOid = principal_key->keyInfo.databaseId;
-		rloc.spcOid = DEFAULTTABLESPACE_OID;
+		rloc.spcOid = DEFAULTTABLESPACE_OID; /* TODO. Revisit*/
```
dAdAbird, line 702: we need to test key rotation in non-default table space.

dAdAbird (approved):

PR #327, opened 2024-11-05
PG-1195 Fix macos CI

artemgavrilov:


ImTheKai (approved):
Tested and worked on my Mac now again as well.

PR #328, opened 2024-11-06
PG-1116 Documented how to define tde_heap as default table access method

nastena1606:
        new file:   documentation/docs/table-access-method.md
        modified:   documentation/mkdocs.yml


documentation/docs/table-access-method.md
```diff
@@ -0,0 +1,85 @@
...
+
+## `tde_heap` and `tde_heap_basic` access methods
+
+The `tde_heap_basic` and `tde_heap` are custom table access methods that come with the `pg_tde` extension to provide data encryption. They are automatically created **only** for the databases where you [enabled the `pg_tde` extension](setup.md) and configured the key provider.
+
+!!! note
+
+    While the `tde_heap_basic` access method is available with both PostgreSQL Community and Percona Server for PostgreSQL, the `tde_heap` (tech preview) table access method is available only with Percona Server for PostgreSQL. Check how to install it [On Ubuntu](apt.md#__tabbed_2_2) and [on RHEL and derivatives](yum.md#__tabbed_2_2)
+
+
```
nastena1606, line 41: Does it make sense to explain how tde_heap works with data? I need your input here, please

ImTheKai:
This PR needs a rebase

nastena1606:
Rebased. I have a pending question in comments. I believe it is useful to explain what happens behind the scenes. I'd like to hear your opinion.

documentation/docs/table-access-method.md
ImTheKai: Yes I think it would be helpful to do that, let me get back to you with a graphic that we can add.

ImTheKai (approved):

documentation/docs/table-access-method.md
```diff
@@ -0,0 +1,96 @@
...
+## Custom access method
+
+You can create a custom table access method for each table and instruct PostgreSQL how to store the data for you. For example, you can tailor the table access method to better suit your specific workload or data access patterns.
+
+To define an access method, use the `CREATE ACCESS METHOD` with the `TYPE` clause set to `table`:
+
+```sql
+CREATE ACCESS METHOD access_method_name TYPE table;
+```
```
artemgavrilov, line 21: Not sure that this section is needed. Probably better will be mentions that different access methods exists and can be installed as an extensions.

documentation/docs/table-access-method.md
```diff
@@ -0,0 +1,96 @@
...
+### How does it work? 
+
+The `tde_heap` access method works on top of the default heap access method and is a marker to point which tables require encryption. It uses the custom storage manager TDE SMGR, which becomes active only after you installed the `pg_tde` extension. 
+
+When a table requires encryption, every data block is encrypted before it is written to disk and decrypted after reading before it is sent to the PostgreSQL core and then to the client. The encryption is done at the storage manager level. 
+
+
+## Changing the default table access method
+
+You can change the default table access method so that every table in the entire database cluster is created using the custom access method. For example, you can enable data encryption by default by defining either `tde_heap_basic` or the  `tde_heap` as the default table access method.
```
artemgavrilov, line 46: We mention `tde_heap_basic` here, but don't provide explanation what it is.

documentation/docs/test.md
```diff
@@ -1,16 +1,14 @@
 # Test Transparent Data Encryption
 
-To check if the data is encrypted, do the following:
+Enabling `pg_tde` extension for a database creates the table access method `tde_heap` . This access method enables you to encrypt the data.
```
artemgavrilov, line 3: And `tde_heap_basic`

nastena1606: I'd stick to having only `tde_heap` AM here. Otherwise it may create a confusion IMHO.

artemgavrilov (approved):

dAdAbird (approved):
LGTM/ Just a minor suggestion

documentation/docs/table-access-method.md
```diff
@@ -0,0 +1,96 @@
...
+    column2 data_type,
+    ...
+) USING tde_heap;
+```
+
+### How `tde_heap` works
+
+The `tde_heap` access method works on top of the default `heap` access method and is a marker to point which tables require encryption. It uses the custom storage manager TDE SMGR, which becomes active only after you installed the `pg_tde` extension. 
+
+WEvery data modification operation is first sent to the Buffer Manager, which updates the buffer cache. Then, it is passed to the storage manager, which then writes it to disk. When a table requires encryption,  the data is sent to the TDE storage manager, where it is encrypted before written to disk.
```
dAdAbird, line 35:
```suggestion
Every data modification operation is first sent to the Buffer Manager, which updates the buffer cache. Then, it is passed to the storage manager, which then writes it to disk. When a table requires encryption, the data is sent to the TDE storage manager, where it is encrypted before written to disk. 
```

PR #329, opened 2024-11-06
Fix CREATE/DROP tde_heap in non-default tablespace

dAdAbird:
1. `tde_mdopen()` is called before `tde_mdcreate()` during CREATE. In the case of non-default tablespace, the creation of the key will fail in tde_mdopen(), as the storage is yet to be created (namely database dir). This doesn't happen with default table space because the db dir had been already created before the CREATE TABLE call.
2. When dropping/moving relations in non-default table space we also have clean-up `*.tde` files. And remove those files physically if there is nothing (tde related) left in the table space. Otherwise `DROP TABLESPACE` will fail.
One caveat is that `tde_mdunlink()` is called after the transaction is commited, so we must not have any errors down the call stack. For that, we need related tdemap* functions to have an option of not failing on error.

Fixes PG-1196

### DIFF
--- a/src/pg_tde_event_capture.c
+++ b/src/pg_tde_event_capture.c
@@ -141,7 +141,7 @@ pg_tde_ddl_command_start_capture(PG_FUNCTION_ARGS)
 		foreach(lcmd, stmt->cmds)
 		{
 			AlterTableCmd *cmd = (AlterTableCmd *) lfirst(lcmd);
-			if (cmd->subtype == AT_SetAccessMethod && 
+			if ((cmd->subtype == AT_SetAccessMethod || cmd->subtype == AT_SetTableSpace) && 
 				((cmd->name != NULL && strcmp(cmd->name, "tde_heap")==0) ||
 				 (cmd->name == NULL && strcmp(default_table_access_method, "tde_heap") == 0))
 				)
--- a/src/smgr/pg_tde_smgr.c
+++ b/src/smgr/pg_tde_smgr.c
@@ -230,8 +230,13 @@ tde_mdreadv(SMgrRelation reln, ForkNumber forknum, BlockNumber blocknum,
 static void
 tde_mdcreate(RelFileLocator relold, SMgrRelation reln, ForkNumber forknum, bool isRedo)
 {
+	RelKeyData* key;
 	TDESMgrRelation tdereln = (TDESMgrRelation) reln;
 
+	/* 
+	 * We have to ensure that all direcories are created before moving on with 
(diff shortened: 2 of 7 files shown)


src/pg_tde_event_capture.c
```diff
@@ -141,7 +141,7 @@ pg_tde_ddl_command_start_capture(PG_FUNCTION_ARGS)
 		foreach(lcmd, stmt->cmds)
 		{
 			AlterTableCmd *cmd = (AlterTableCmd *) lfirst(lcmd);
-			if (cmd->subtype == AT_SetAccessMethod && 
+			if ((cmd->subtype == AT_SetAccessMethod || cmd->subtype == AT_SetTableSpace) &&
```
dutow, line 144: Are you sure about this? I think `name` will refer to the tablespace name for `SET TABLESPACE`, not the access method name, making the next access method name checks incorrect. In case of `SET TABLESPACE`, we should open the table instead and check the current access method.

Also, the TODO comment below is outdated with these changes.

src/smgr/pg_tde_smgr.c
```diff
@@ -68,19 +68,23 @@ tde_smgr_get_key(SMgrRelation reln, RelFileLocator* old_locator)
 		return rkd;
 	}
 
+	if (!createIfNone)
+		return NULL;
+
 	/* if this is a CREATE TABLE, we have to generate the key */
-	if (event->encryptMode == true && event->eventType == TDE_TABLE_CREATE_EVENT)
+	if(event->encryptMode == true && event->eventType == TDE_TABLE_CREATE_EVENT)
```
dutow, line 75: accidental whitespace change?

src/smgr/pg_tde_smgr.c
```diff
@@ -254,14 +260,20 @@ tde_mdcreate(RelFileLocator relold, SMgrRelation reln, ForkNumber forknum, bool
...
 {
 	TDESMgrRelation tdereln = (TDESMgrRelation) reln;
-	RelKeyData *key = tde_smgr_get_key(reln, NULL);
 
+	/* 
+	 * tde_mdopen() is called before tde_mdcreate() during CREATE. And in case
+	 * of non-default tablespace, the creation of the key will fail here, as
+	 * the storage is yet to be created.
+	 */
+	RelKeyData *key = tde_smgr_get_key(reln, NULL, false);
```
dutow, line 275: But we will emit warnings even with `false`, that could be confusing.

dAdAbird:
Will fix in https://perconadev.atlassian.net/browse/PG-1209

PR #333, opened 2024-11-08
PG-1198 Refactor docs and docker workflows

artemgavrilov:
PG-1198

This PR adds human-readable names for docker and documentation workflows/jobs. Two docker related workflows merged into single one and refactored to use official docker actions to build and push.

Screenshot 2024-11-08 at 17 17 15


.github/workflows/docs.yaml
```diff
@@ -1,29 +1,30 @@
...
+        working-directory: ${{ github.workspace }}/documentation
 
-    steps: 
-      #Pull the latest changes
-      - name: Chekout code
+    steps:
+      - name: Chekout
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
```
artemgavrilov, line 21: @nastena1606  Not sure why do we need to fetch all history here instead of fetching just HEAD

nastena1606: I used it the same way it's done for other doc projects. If you feel there's the better way, go ahead.

ImTheKai (approved):

nastena1606 (approved):

PR #339, opened 2024-11-12
PG-1126 fix: only check API version with our fork

dutow:


dAdAbird (approved):
Oops, that thought came to me after I reviewed the prev PR and then I got distracted and forgot...

PR #357, opened 2024-11-22
PG-1120: Add a new SQL interface pg_tde_alter_principal_key_keyring

codeforall:
Description
Add a new SQL interface pg_tde_alter_principal_key_keyring that allows the user to change the key provider for the principal key without changing or rotating the actual key

Links
https://perconadev.atlassian.net/browse/PG-1120


ImTheKai:
> Tap tests has to be added into season builds but overall LGTM

Will you take care of that in a new PR?

PR #368, opened 2024-12-04
PG-1241 Documented KMIP integration and setup

nastena1606:
### Description
Documented KMIP integration and setup

modified:   documentation/docs/_images/tde-flow.png
        modified:   documentation/docs/tde.md


documentation/docs/setup.md
```diff
@@ -42,11 +42,37 @@ Load the `pg_tde` at the start time. The extension requires additional shared me
...
 
-1. Set up a key provider for the database where you have enabled the extension
+1. Set up a key provider for the database where you have enabled the extension.
 
-    === "With HashiCorp Vault"
+    === "With KMIP server"
+
+        Make sure you have obtained the root certificate for the KMIP server and the keypair for the client. The client key needs permissions to create / read keys on the server.
+        
+        For testing purposes you can use the OpenSSL to issue self-signed certificates. For production use we recommend you to use the valid certificates issued by the key management appliance.
```
dutow, line 51: "For testing purposes you can use the OpenSSL to issue self-signed certificates. " - I don't understand this part. For testing with a real server, you need proper certificates. For tetsing with the pykmip dummy server, it has scripts to setup certificates.

documentation/docs/setup.md
```diff
@@ -42,11 +42,37 @@ Load the `pg_tde` at the start time. The extension requires additional shared me
 
 ## Key provider configuration
 
-1. Set up a key provider for the database where you have enabled the extension
+1. Set up a key provider for the database where you have enabled the extension.
 
-    === "With HashiCorp Vault"
+    === "With KMIP server"
+
+        Make sure you have obtained the root certificate for the KMIP server and the keypair for the client. The client key needs permissions to create / read keys on the server.
```
dutow, line 49: You only need the private key for the client for connection, the public key is stored on the server. It also need to contain the certificate chain for the key.

Root certificate can be technically empty, if the machine running postgres trusts it.

documentation/docs/setup.md
```diff
@@ -42,11 +42,37 @@ Load the `pg_tde` at the start time. The extension requires additional shared me
...
+
+        ```
+        SELECT pg_tde_add_key_provider_kmip('provider-name','kmip-IP', 5696, '/path_to/server_certificate.pem', '/path_to/client_key.pem');
+        ```
+
+        where:
+
+        * `provider-name` is the name of the provider. You can specify any name, it's for you to identify the provider.
+        * `kmip-IP` is the IP address of a domain name of the KMIP server
+        * The port to communicate with the KMIP server. When undefined, the default port 5696 is used
```
dutow, line 61: Currently it is a required parameter, but making it optional is a good idea.

documentation/docs/setup.md
```diff
@@ -57,78 +83,77 @@ Load the `pg_tde` at the start time. The extension requires additional shared me
         * `secret_token` is an access token with read and write access to the above mount point
         * [optional] `ca_path` is the path of the CA file used for SSL verification
 
+        <i warning>:material-information: Warning:</i> This example is for testing purposes only:
 
-    === "With keyring file"
+	    ```sql
+	    SELECT pg_tde_add_key_provider_file('file-vault','/tmp/pg_tde_test_local_keyring.per');
```
dutow, line 89: This is an example for keyring file, not vault.

documentation/docs/setup.md
nastena1606: So to rewrite it, can we say "for testing purposes use the pykmip dummy server which enables you to set up certificates. For using the KMIP server, make sure to obtain valid certificates issued by the key management appliance." ?

documentation/docs/setup.md
nastena1606: How to get the private key for the client?

nastena1606: I'll keep this info and give the link to the Vault tutorial

documentation/docs/setup.md
nastena1606: Please help with the example for vault. For users to have an idea

dutow:
```
SELECT pg_tde_add_key_provider_vault_v2('name','root_token','http://127.0.0.1:8200','secret',NULL);
```

name, root token (this is the "password"), url, name of the mount point to use on vault, and an optional CA file to verify the SSL connection.

nastena1606: Thank, updated

documentation/docs/setup.md
nastena1606: Updated

documentation/docs/functions.md
```diff
@@ -33,6 +33,24 @@ where:
...
+
+```
+SELECT pg_tde_add_key_provider_kmip('provider-name','kmip-IP', 5696, '/path_to/server_certificate.pem', '/path_to/client_key.pem');
+```
+
+where:
+
+* `provider-name` is the name of the provider. You can specify any name, it's for you to identify the provider.
+* `kmip-IP` is the IP address of a domain name of the KMIP server
+* The port to communicate with the KMIP server. The default port is `5696`.
```
dutow, line 50: We still don't have a default port, it is a required parameter.

documentation/docs/tde.md
```diff
@@ -6,7 +6,7 @@ Transparent Data Encryption is a technology to protect data at rest. The encrypt
 
 To encrypt the data, two types of keys are used:
 
-* Database keys to encrypt user data. These are stored internally, near the data that they encrypt.
+* Table encryption keys (TEK) to encrypt user data. These keys are stored internally, near the data that they encrypt.
```
dutow, line 9: Do we need this acronym? We never used that before (TEK).

documentation/docs/setup.md
```diff
@@ -58,12 +58,14 @@ Load the `pg_tde` at the start time. The extension requires additional shared me
 
         * `provider-name` is the name of the provider. You can specify any name, it's for you to identify the provider.
         * `kmip-IP` is the IP address of a domain name of the KMIP server
-        * The port to communicate with the KMIP server. The default port is `5696`.
+        * `5696` is the port to communicate with the KMIP server.
```
dutow, line 61: `port` is the port to communicate with the KMIP server? It can be anything, depending on the KMIP server, 5696 is just the official port mot servers use...

documentation/docs/tde.md
nastena1606: It's been used in Jan's talks. But TBH I like database keys more)
Changing back

documentation/docs/setup.md
nastena1606: But it is hardcoded now? So not a default one, but required one, correct? Why does the current wording not suit?

documentation/docs/tde.md
dutow: Both are incorrect. we are simply calling them internal keys in the code, because that's what they are.

These can encrypt in theory any database file - tables, indexes, WAl, who knows what in the future...

documentation/docs/setup.md
dutow: No, it is a parameter the user has to specify.

nastena1606: Updated

documentation/docs/tde.md
nastena1606: Updated

dutow (approved):

PR #390, opened 2024-12-17
Documentation github action fix: mike order and check depth

dutow:
PG-0

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Chekout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
     steps:
       - name: Chekout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0 # fetch all commits/branches
```
artemgavrilov, line 25: Aha, so that was intentionally. My bad.

dutow: Yes, without this the checkout creates a shallow clone, and `mike deploy` fails because it creates a new local `gh-pages` branch istead of updating the already exisitng branch

PR #391, opened 2024-12-20
PG-1268 Added use cases as FAQ

nastena1606:
PG-1268

### Description
Added most common uses cases, best practices in the form of an FAQ page

modified:   documentation/docs/faq.md

### DIFF
--- a/documentation/docs/faq.md
+++ b/documentation/docs/faq.md
@@ -81,9 +81,9 @@ The initial decision on what file to encrypt is based on the table access method
 
 The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
 
-WAL encryption is done globally for the entire using the principal key. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
+WAL encryption is done globally for the entire database cluster using the global principal key. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
 
-You can turn WAL encryption on and off so WAL can contain both encrypted and unencrypted pages. The encryption influences only writes.
+You can turn WAL encryption on and off so WAL can contain both encrypted and unencrypted pages. The WAL encryption GUC variable influences only writes.
 
 Whenever the WAL is being read (by the recovery process or tools), the decision on what pages should be decrypted is based solely on the encryption flag of each page.
 
--- a/documentation/docs/faq.md
+++ b/documentation/docs/faq.md
@@ -25,7 +25,7 @@ Using TDE helps you avoid the following risks:
 
 If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
 
-`pg_tde` does not encrypt system catalogs yet. This is planned for future releases. 
+`pg_tde` does not encrypt system catalogs yet. This means that statistics data and database metadata are not encrypted. The encryption of system catalogs is planned for future releases. 
 
 
 ## I use disk-level encryption. Why should I care about TDE?
@@ -41,6 +41,8 @@ Another point to consider is PCI DSS compliance for Personal Account Numbers (PA
    * Separate the logical data access from the operating system authentication.
 
    * Ensure the decryption key is not linked to user accounts.
+
+   Note that PCI DSS 3.4.1 is retiring on March 31, 2025. Therefore, consider switching to PCI DSS 4.0.
 
 * **PCI DSS 4.0** standards consider using only disk and partition-level encryption not enough to ensure PAN protection. It requires an additional layer of security that `pg_tde` can provide. 
 
@@ -81,7 +83,11 @@ The initial decision on what file to encrypt is based on the table access method
 
 The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
 
-WAL encryption is done globally for the entire database cluster using the global internal and principal keys. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
+WAL encryption is done globally for the entire database cluster. All modifications to any database within a PostgreSQL cluster are written to the same WAL to maintain data consistency and integrity and ensure that PostgreSQL cluster can be restored to a consistent state. Therefore, WAL is encrypted globally. 
+
+When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
+
+The same 2-tier approach is used with WAL as with the table data: WAL pages are first encrypted with the internal key. Then the internal key is encrypted with the global principal key.
 
 You can turn WAL encryption on and off so WAL can contain both encrypted and unencrypted pages. The WAL encryption GUC variable influences only writes.
 
--- a/documentation/docs/faq.md
+++ b/documentation/docs/faq.md
@@ -25,12 +25,14 @@ Using TDE helps you avoid the following risks:
 
 If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
 
+`pg_tde` does not encrypt system catalogs yet. This is planned for future releases. 
+
 
 ## I use disk-level encryption. Why should I care about TDE?
 
 Encrypting a hard drive encrypts all data, including system, application, and temporary files.
 
-Full disk encryption protects your data from people who have physical access to your device and even if it is lost or stolen. However, it doesn't protect the data after system boot-up: the data is automatically decrypted when the system runs or when an authorized user requests it. This means a sophisticated attacker can access the memory and steal the encryption keys through a cold boot attack.
+Full disk encryption protects your data from people who have physical access to your device and even if it is lost or stolen. However, it doesn't protect the data after system boot-up: the data is automatically decrypted when the system runs or when an authorized user requests it. 
 
 Another point to consider is PCI DSS compliance for Personal Account Numbers (PAN) encryption.
 
@@ -40,9 +42,9 @@ Another point to consider is PCI DSS compliance for Personal Account Numbers (PA
 
    * Ensure the decryption key is not linked to user accounts.
 
-* **PCI DSS 4.0** standards consider using only disk and partition-level encryption not enough to ensure PAN protection. It requires an additional layer of security that TDE can provide. 
+* **PCI DSS 4.0** standards consider using only disk and partition-level encryption not enough to ensure PAN protection. It requires an additional layer of security that `pg_tde` can provide. 
 
-TDE focuses specifically on data files and offers more granular control over encrypted data. It also ensures that files are encrypted on disk during runtime and when moved to another system or storage.
+`pg_tde` focuses specifically on data files and offers more granular control over encrypted data. The data remains encrypted on disk during runtime and when you move it to another directory, another system or storage. An example of such data is backups. They remain encrypted when moved to the backup storage.
 
 Thus, to protect your sensitive data, consider using TDE to encrypt it at the table level. Then use disk-level encryption to encrypt a specific volume where this data is stored, or the entire disk.
 
@@ -75,7 +77,7 @@ Here’s how encryption works:
 
 First, data files are encrypted with internal keys. Each file that has a different OID, has an internal key. For example, a table with 4 indexes will have 5 internal keys - one for the table and one for each index.	
 
-The initial decision on what file to encrypt is based on the PostgreSQL triggers. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
+The initial decision on what file to encrypt is based on the table access method in PostgreSQL. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
 
 The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
 
@@ -104,14 +106,19 @@ No, it's not yet supported. In our implementation we reply on OpenSSL libraries 
 Yes, you can encrypt an existing table. Run the ALTER TABLE command as follows:
 
 ```
-ALTER TABLE table_name SET access method tde_heap;
+ALTER TABLE table_name SET ACCESS METHOD tde_heap;
 ```
 
-Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the SELECT COUNT(*) command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section. 
+Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the `SELECT COUNT(*)` command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section. 
 
 ## Do I have to restart the database to encrypt the data?
 
-No, you don't have to restart the database to encrypt the data. When you create or alter the table using the `tde_heap` access method, the files are marked as those that require encryption. The encryption happens at the storage manager level, before a transaction is written to disk. Read more about how `tde_heap` access method works in the [How tde_heap works](table-access-method.md#how-tde_heap-works) section.
+You must restart the database in the following cases to apply the changes:
+
+* after you enabled the `pg_tde` extension
+* to tun on / off the WAL encryption
+
+After that, no database restart is required. When you create or alter the table using the `tde_heap` access method, the files are marked as those that require encryption. The encryption happens at the storage manager level, before a transaction is written to disk. Read more about [how tde_heap works](table-access-method.md#how-tde_heap-works).
 
 ## What happens to my data if I lose a principal key?
 
@@ -124,6 +131,8 @@ Multi-tenancy is the type of architecture where multiple users, or tenants, shar
 In `pg_tde`, multi-tenancy is supported via a separate principal key per database. This means that a database owner can decide what tables to encrypt within a database. The same database can have both encrypted and non-encrypted tables.
 
 To control user access to the databases, you can use role-based access control (RBAC).
+
+WAL files are encrypted globally across the entire PostgreSQL cluster using the same encryption key. Users don't interact with WAL files as these are used by the database management system to ensure data integrity and durability.
 
 ## Are my backups safe? Can I restore from them?
 
--- a/documentation/docs/faq.md
+++ b/documentation/docs/faq.md
@@ -2,10 +2,29 @@
 
 ## Why do I need TDE?
 
-- Compliance to security and legal regulations like GDPR, PCI DSS and others
-- Encryption of backups
-- Granular encryption of specific data sets and reducing the performance overhead that encryption brings
+- Compliance to security and legal regulations like General Data Protection Regulation (GDPR), Payment Card Industry Data Security Standard (PCI DSS), California Consumer Privacy Act (CCPA), Data Protection Act 2018 (DPA 2018) and others
+- Encryption of backups. Even when an authorized person gets physical access to a backup, encryption ensures that the data remains unreadable and secure.
+- Granular encryption of specific data sets and reducing the performance overhead that encryption brings. 
 - Additional layer of security to existing security measures
+
+## When and how should I use TDE?
+
+If you are dealing with Personally Identifiable Information (PII), data encryption is crucial. Especially if you are involved in areas like with strict regulations like:
+
+* financial services where TDE helps to comply with PCI DSS, 
+* healthcare and insurance - compliance with HIPAA, 
+* telecommunications, government and education to ensure data confidentiality.
+
+Using TDE helps you avoid the following risks:
+
+* Data breaches
+* Identity theft that may lead to financial fraud and other crimes
+* Reputation damage leading to loss of customer trust and business
+* Legal consequences and financial losses for non-compliance with data protection regulations
+* Internal threats by misusing unencrypted sensitive data 
+
+If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
+
 
 ## I use disk-level encryption. Why should I care about TDE?
 
@@ -13,7 +32,7 @@ Encrypting a hard drive encrypts all data including system and application files
 
 TDE focuses specifically on data files and offers a more granular control over encrypted data. It also ensures that files are encrypted on disk during runtime and when moved to another system or storage.
 
-Consider using TDE and storage-level encryption together to add another layer of data security
+Consider using TDE and storage-level encryption together to add another layer of data security.
 
 ## Is TDE enough to ensure data security?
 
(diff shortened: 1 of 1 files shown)


documentation/docs/faq.md
```diff
@@ -2,18 +2,37 @@
...
-- Encryption of backups
-- Granular encryption of specific data sets and reducing the performance overhead that encryption brings
+- Compliance to security and legal regulations like General Data Protection Regulation (GDPR), Payment Card Industry Data Security Standard (PCI DSS), California Consumer Privacy Act (CCPA), Data Protection Act 2018 (DPA 2018) and others
+- Encryption of backups. Even when an authorized person gets physical access to a backup, encryption ensures that the data remains unreadable and secure.
+- Granular encryption of specific data sets and reducing the performance overhead that encryption brings. 
 - Additional layer of security to existing security measures
 
+## When and how should I use TDE?
+
+If you are dealing with Personally Identifiable Information (PII), data encryption is crucial. Especially if you are involved in areas like with strict regulations like:
```
ImTheKai, line 12: ..in areas with strict regulations... -> "with" can be dropped here

ImTheKai (approved):
LGTM, just one minor word

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+## Do I have to restart the database to encrypt the data?
+
+No, you don't have to restart the database to encrypt the data. When you create or alter the table using the `tde_heap` access method, the files are marked as those that require encryption. The encryption happens at the storage manager level, before a transaction is written to disk. Read more about how `tde_heap` access method works in the [How tde_heap works](table-access-method.md#how-tde_heap-works) section.
+
 ## What happens to my data if I lose a principal key?
 
-If you lose encryption keys, especially, the principal key, the data is lost. That's why it's critical to back up your encryption keys securely.
\ No newline at end of file
+If you lose encryption keys, especially, the principal key, the data is lost. That's why it's critical to back up your encryption keys securely.
```
janwieremjewicz, line 96: KMS usage is also a quite important principle to leverage for this particular reason!

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+## Can I use `pg_tde` in a multi-tenant setup?
+
+Multi-tenancy is the type of architecture where multiple users, or tenants, share the same resource. It can be a database, a schema or an entire cluster. 
+
+In `pg_tde`, multi-tenancy is supported via a separate principal key per database. This means that a database owner can decide what tables to encrypt within a database. The same database can have both encrypted and non-encrypted tables.
```
janwieremjewicz, line 102: AND each database owner can do this on their own within the same PostgreSQL installation (or instance) known as a PostgreSQL cluster in the PG-specific lingo

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+To control user access to the databases, you can use role-based access control (RBAC).
+
+## Are my backups safe? Can I restore from them?
+
+`pg_tde` encrypts data at rest. This means that data is stored on disk in an encrypted form. During a backup, already encrypted data files are copied from disk onto the storage. This ensures the data safety in backups.
+
+Since the encryption happens on the database level, it makes no difference for your tools and applications. They work with the data in the same way.
+
+To restore from an encrypted backup, you must have the same principal encryption key, which was used to encrypt files in your backup.  
+
```
janwieremjewicz, line 113: I think that this is crucial to mention that if changes have happened with KMS since backup was made, some extra reconfig for the backup will be needed

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+## I'm using the FIPS mode. Am I safe to use it and `pg_tde`? Can I use my own OpenSSL library in the FIPS mode and `pg_tde` together?
```
janwieremjewicz, line 114: We need to clarify this entry more. 
FIPS mode is an option Open-SSL has. For PostgreSQL I think that this is a [very good write-up](https://peter.eisentraut.org/blog/2023/12/05/postgresql-and-fips-mode) for pg_tde as we rely on OpenSSL provided by the user. Whether they they use [it as FIPS validated](https://github.com/openssl/openssl/blob/master/README-FIPS.md) and how they do is up to them.

janwieremjewicz (commented):
I've left a bunch of comments that should be addressed before it's ready.

documentation/docs/faq.md
```diff
@@ -2,18 +2,37 @@
...
+* Data breaches
+* Identity theft that may lead to financial fraud and other crimes
+* Reputation damage leading to loss of customer trust and business
+* Legal consequences and financial losses for non-compliance with data protection regulations
+* Internal threats by misusing unencrypted sensitive data 
+
+If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
+
+
 ## I use disk-level encryption. Why should I care about TDE?
```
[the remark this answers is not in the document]
nastena1606: Updated

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+The initial decision on what file to encrypt is based on the PostgreSQL triggers. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
+
+The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
+
+## Should I encrypt all my data?
+
+It depends on your business requirements and the sensitivity of your data. Encrypting all data is a good practice but it can have a performance impact. 
+
+Consider encrypting only tables that store sensitive data. `pg_tde` supports multi-tenancy enabling you to do just that. You can decide what tables to encrypt and with what key. The [Setup](setup.md) section in documentation focuses on this approach.
```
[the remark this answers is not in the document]
nastena1606: I've excluded mentioning multi-tenancy support from this answer

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
 * Regular monitoring and auditing
 * Additional data protection for sensitive fields (e.g., application-layer encryption)
 
+## How does `pg_tde` make my data safe?
+
+`pg_tde` uses two keys to encrypt data:
+
+* Internal encryption keys to encrypt the data. These keys are stored internally, in a single `$PGDATA/pg_tde` directory.
+* Principal keys to encrypt table encryption keys. These keys are stored externally, in the Key Management Store (KMS). You can use either the HashiCorp Vault server or the KMIP-compatible server.
```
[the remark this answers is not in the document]
nastena1606: updated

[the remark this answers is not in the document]
nastena1606: Updated

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+`pg_tde` currently uses a AES-CBC-128 algorithm. First the internal keys in the datafile are encrypted using the principal key with AES-CBC-128, then the file data itself is again encrypted using AES-CBC-128 with the internal key.
+
+For WAL encryption, AES-CTR-128 is used.
+
+The support of other encryption mechanisms such as AES256 is planned for future releases.
+
+## Is post-quantum encryption supported?
+
+No, it's not yet supported.
```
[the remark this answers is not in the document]
nastena1606: Updated with a brief reasoning why

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+## Can I encrypt an existing table?
+
+Yes, you can encrypt an existing table. Run the ALTER TABLE command as follows:
```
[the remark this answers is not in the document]
nastena1606: Added

documentation/docs/faq.md
```diff
@@ -2,18 +2,37 @@
...
+- Encryption of backups. Even when an authorized person gets physical access to a backup, encryption ensures that the data remains unreadable and secure.
+- Granular encryption of specific data sets and reducing the performance overhead that encryption brings. 
 - Additional layer of security to existing security measures
 
+## When and how should I use TDE?
+
+If you are dealing with Personally Identifiable Information (PII), data encryption is crucial. Especially if you are involved in areas like with strict regulations like:
+
+* financial services where TDE helps to comply with PCI DSS, 
+* healthcare and insurance - compliance with HIPAA,
```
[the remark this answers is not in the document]
nastena1606: Mentioned

documentation/docs/faq.md
```diff
@@ -25,6 +44,73 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+We advise encrypting the whole database only if all your data is sensitive, like PII, or if there is no other way to comply with data safety requirements. See [How to configure global encryption](global-encryption.md).
+
+## What cipher mechanisms are used by `pg_tde`?
+
+`pg_tde` currently uses a AES-CBC-128 algorithm. First the internal keys in the datafile are encrypted using the principal key with AES-CBC-128, then the file data itself is again encrypted using AES-CBC-128 with the internal key.
+
+For WAL encryption, AES-CTR-128 is used.
+
+The support of other encryption mechanisms such as AES256 is planned for future releases.
```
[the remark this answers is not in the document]
nastena1606: Updated

documentation/docs/faq.md
nastena1606: I'm sorry I don't understand the point here. That a db owner can manage own DB without the involvement of the DB admin?

documentation/docs/faq.md
nastena1606: Could you please elaborate what I should clarify here? According to the blog post PostgreSQL ≥17: all PostgreSQL test suites pass with OpenSSL in FIPS mode and since pg_tde is for PG 17, the user is fine. 
Also, according to @dutow pg_tde uses the same OpenSSL as PostgreSQL does. So if PG uses OpenSSL with FIPS mode enabled, so does pg_tde. 
Or am I missing something?

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+## Is post-quantum encryption supported?
+
+No, it's not yet supported. In our implementation we reply on OpenSSL libraries that don't yet support post-quantum encryption.
+
+## Can I encrypt an existing table?
+
+Yes, you can encrypt an existing table. Run the ALTER TABLE command as follows:
+
+```
+ALTER TABLE table_name SET access method tde_heap;
```
dutow, line 107: `ACCESS METHOD`

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+```
+
+Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the SELECT COUNT(*) command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section.
```
dutow, line 110: `SELECT COUNT(*)

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+```
+ALTER TABLE table_name SET access method tde_heap;
+```
+
+Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the SELECT COUNT(*) command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section. 
+
+## Do I have to restart the database to encrypt the data?
+
+No, you don't have to restart the database to encrypt the data. When you create or alter the table using the `tde_heap` access method, the files are marked as those that require encryption. The encryption happens at the storage manager level, before a transaction is written to disk. Read more about how `tde_heap` access method works in the [How tde_heap works](table-access-method.md#how-tde_heap-works) section.
```
dutow, line 114: You have to restart the database to enable pg_tde, or to turn wal encryption on/off.

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
 
-If you lose encryption keys, especially, the principal key, the data is lost. That's why it's critical to back up your encryption keys securely.
\ No newline at end of file
+If you lose encryption keys, especially, the principal key, the data is lost. That's why it's critical to back up your encryption keys securely and use the Key Management service for key management.
+
+## Can I use `pg_tde` in a multi-tenant setup?
+
+Multi-tenancy is the type of architecture where multiple users, or tenants, share the same resource. It can be a database, a schema or an entire cluster. 
+
+In `pg_tde`, multi-tenancy is supported via a separate principal key per database. This means that a database owner can decide what tables to encrypt within a database. The same database can have both encrypted and non-encrypted tables.
```
dutow, line 124: The WAL is still encrypted globally with the same key for everyone.

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+To control user access to the databases, you can use role-based access control (RBAC).
+
+## Are my backups safe? Can I restore from them?
+
+`pg_tde` encrypts data at rest. This means that data is stored on disk in an encrypted form. During a backup, already encrypted data files are copied from disk onto the storage. This ensures the data safety in backups.
+
+Since the encryption happens on the database level, it makes no difference for your tools and applications. They work with the data in the same way.
```
dutow, line 132: I assume you mean "backup tools" here, not any command line pg tools?

We actually have a few issues with backup tools, because we add additional files, and checksums also work differently (currently not, without modifications to the tools)

documentation/docs/faq.md
```diff
@@ -2,18 +2,49 @@
...
+
+Using TDE helps you avoid the following risks:
+
+* Data breaches
+* Identity theft that may lead to financial fraud and other crimes
+* Reputation damage leading to loss of customer trust and business
+* Legal consequences and financial losses for non-compliance with data protection regulations
+* Internal threats by misusing unencrypted sensitive data 
+
+If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
```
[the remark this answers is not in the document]
nastena1606: So should I mention we don't encrypt system catalogs yet but plan to do so?

documentation/docs/faq.md
```diff
@@ -2,18 +2,49 @@
...
+
+
 ## I use disk-level encryption. Why should I care about TDE?
 
-Encrypting a hard drive encrypts all data including system and application files that are there. However, disk encryption doesn’t protect your data after the boot-up of your system. During runtime, the files are decrypted with disk-encryption.
+Encrypting a hard drive encrypts all data, including system, application, and temporary files.
+
+Full disk encryption protects your data from people who have physical access to your device and even if it is lost or stolen. However, it doesn't protect the data after system boot-up: the data is automatically decrypted when the system runs or when an authorized user requests it. This means a sophisticated attacker can access the memory and steal the encryption keys through a cold boot attack.
```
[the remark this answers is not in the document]
nastena1606: But do we store encryption keys for TDE in memory? 
What will be a better way to promote tde here, to explain users why they need it?

documentation/docs/faq.md
```diff
@@ -25,6 +56,83 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+* [OpenBao](https://openbao.org/) implementation of Vault
+* KMIP-compatible server. KMIP is a standardized protocol for handling cryptographic workloads and secrets management
+
+HashiCorp Vault can also act as the KMIP server, managing cryptographic keys for clients that use the KMIP protocol. 
+
+Here’s how encryption works:
+
+First, data files are encrypted with internal keys. Each file that has a different OID, has an internal key. For example, a table with 4 indexes will have 5 internal keys - one for the table and one for each index.	
+
+The initial decision on what file to encrypt is based on the PostgreSQL triggers. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not.
```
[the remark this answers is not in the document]
nastena1606: Updated

documentation/docs/faq.md
nastena1606: Updated

documentation/docs/faq.md
nastena1606: So what will be the better way to answer the question? What should users know, do to ensure their backup tools work without disrupition?

documentation/docs/faq.md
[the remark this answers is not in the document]
dutow: If we want to say that we encrypt everything then yes, because it's only 99% true.

documentation/docs/faq.md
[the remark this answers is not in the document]
dutow: Yes, we store the keys and the decrypted data both in memory.

The only difference between full disk encryption and tde is that tde not tied to the disk/OS: if you move the datadir to a different dir, it remains encrypted (e.g. backups). And it isn't accessible automatically after system start, it needs another form of authentication (to the KMS where the principal keys are stored)

documentation/docs/faq.md
dutow: I think Shahid is testing one of the backup tools currently, so that's still an active question. Generally we know that checksums do not work, otherwise we currently think that backup tools work correctly.

documentation/docs/faq.md
```diff
@@ -104,14 +106,19 @@ No, it's not yet supported. In our implementation we reply on OpenSSL libraries
...
-Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the SELECT COUNT(*) command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section. 
+Since the `ALTER TABLE SET` command drops hint bits and this may affect the performance, we recommend to run the `SELECT COUNT(*)` command. It checks every tuple for visibility and sets its hint bits. Read more in the [Changing existing table](test.md) section. 
 
 ## Do I have to restart the database to encrypt the data?
 
-No, you don't have to restart the database to encrypt the data. When you create or alter the table using the `tde_heap` access method, the files are marked as those that require encryption. The encryption happens at the storage manager level, before a transaction is written to disk. Read more about how `tde_heap` access method works in the [How tde_heap works](table-access-method.md#how-tde_heap-works) section.
+You must restart the database in the following cases to apply the changes:
+
+* after you enabled the `pg_tde` extension
+* to tun on / off the WAL encryption
```
dutow, line 119: typo: turn

dutow (approved):

documentation/docs/faq.md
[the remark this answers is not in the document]
jeltz: Are temporary files really encrypted? Our README says they are not.

documentation/docs/faq.md
```diff
@@ -2,18 +2,51 @@
...
+
+`pg_tde` does not encrypt system catalogs yet. This is planned for future releases. 
+
+
 ## I use disk-level encryption. Why should I care about TDE?
 
-Encrypting a hard drive encrypts all data including system and application files that are there. However, disk encryption doesn’t protect your data after the boot-up of your system. During runtime, the files are decrypted with disk-encryption.
+Encrypting a hard drive encrypts all data, including system, application, and temporary files.
+
+Full disk encryption protects your data from people who have physical access to your device and even if it is lost or stolen. However, it doesn't protect the data after system boot-up: the data is automatically decrypted when the system runs or when an authorized user requests it.
```
jeltz, line 35: That data is autoamtically encrypted is not necessarily true. You might have a local key server or have to have someone have to enter a password to decrypt the disk.

documentation/docs/faq.md
```diff
@@ -25,6 +58,90 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+HashiCorp Vault can also act as the KMIP server, managing cryptographic keys for clients that use the KMIP protocol. 
+
+Here’s how encryption works:
+
+First, data files are encrypted with internal keys. Each file that has a different OID, has an internal key. For example, a table with 4 indexes will have 5 internal keys - one for the table and one for each index.	
+
+The initial decision on what file to encrypt is based on the table access method in PostgreSQL. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
+
+The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
+
```
jeltz, line 83: We should mentioned something about WAL here.

documentation/docs/faq.md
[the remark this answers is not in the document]
nastena1606: @dutow please confirm, should I remove temporary files from this part and the docs: https://percona.github.io/pg_tde/main/#whats-encrypted ?

documentation/docs/faq.md
nastena1606: Doesn't it cover the case "when an authorized user requests it (to decrypt the data)". ?

documentation/docs/faq.md
nastena1606: Added info from architecture doc

documentation/docs/faq.md
dutow: I agree, if you need a key server or manual password, that's *before* the system runs.

documentation/docs/faq.md
```diff
@@ -73,14 +73,21 @@ You can use the following KMSs:
...
 
 First, data files are encrypted with internal keys. Each file that has a different OID, has an internal key. For example, a table with 4 indexes will have 5 internal keys - one for the table and one for each index.	
 
 The initial decision on what file to encrypt is based on the table access method in PostgreSQL. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
 
 The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
 
+WAL encryption is done globally for the entire using the principal key. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not.
```
dutow, line 84: "for the entire server/cluster/database (something)"

"Using the global/server principal key"

documentation/docs/faq.md
```diff
@@ -73,14 +73,21 @@ You can use the following KMSs:
...
 
 The initial decision on what file to encrypt is based on the table access method in PostgreSQL. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
 
 The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
 
+WAL encryption is done globally for the entire using the principal key. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
+
+You can turn WAL encryption on and off so WAL can contain both encrypted and unencrypted pages. The encryption influences only writes.
```
dutow, line 86: "The WAL encryption GUC variable" influences only writes - the encryption itself of course influences the pages that were already encrypted :)

documentation/docs/faq.md
```diff
@@ -25,6 +58,97 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
 * Regular monitoring and auditing
 * Additional data protection for sensitive fields (e.g., application-layer encryption)
 
+## How does `pg_tde` make my data safe?
+
+`pg_tde` uses two keys to encrypt data:
+
+* Internal encryption keys to encrypt the data. These keys are stored internally, in a single `$PGDATA/pg_tde` directory.
```
dAdAbird, line 65: Maybe to write here that "These keys are stored **_encrypted_** internally..."

documentation/docs/faq.md
```diff
@@ -25,6 +58,97 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+* Principal keys to encrypt table encryption keys. These keys are stored externally, in the Key Management System (KMS).
```
dAdAbird, line 66: I thinks either we should use "internal keys" here or "table encryption keys" in the previous point not to confuse the reader. I think "internal keys" better as we use it not only for table data technically speaking.

documentation/docs/faq.md
```diff
@@ -25,6 +58,97 @@ No. TDE is an additional layer to ensure data security. It protects data at rest
...
+
+The initial decision on what file to encrypt is based on the table access method in PostgreSQL. When you run a `CREATE` or `ALTER TABLE` statement with the `USING tde_heap` clause, the newly created data files are marked as encrypted, and then file operations encrypt/decrypt the data. Later, if an initial file is re-created as a result of a `TRUNCATE` or `VACUUM FULL` command, the newly created file inherits the encryption information and is either encrypted or not. 
+
+The principal key is used to encrypt the internal keys. The principal key is stored in the key management store. When you query the table, the principal key is retrieved from the key store to decrypt the table. Then the internal key for that table is used to decrypt the data.
+
+WAL encryption is done globally for the entire database cluster using the global principal key. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not.
```
dAdAbird, line 84:
```suggestion
WAL encryption is done globally for the entire database cluster using the global internal and principal keys. When you turn on WAL encryption, `pg_tde` encrypts entire WAL pages except for the header. The header contains a marker if a page is encrypted or not. 
```

nastena1606: @dAdAbird could you please add how WAL encryption works? I mean in the way data encryption is described, what is encrypted with internal keys, and what with external. I believe this will make things clearer.

documentation/docs/faq.md
nastena1606: Fixed

documentation/docs/faq.md
nastena1606: Added

documentation/docs/faq.md
nastena1606: Added

documentation/docs/faq.md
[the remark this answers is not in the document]
nastena1606: Changed temporary files to temporary tables

documentation/docs/faq.md
```diff
@@ -28,11 +28,23 @@ If to translate sensitive data to files stored in your database, these are user
...
 
-Consider using TDE and storage-level encryption together to add another layer of data security.
+Another point to consider is PCI DSS compliance for Personal Account Numbers (PAN) encryption.
+
+* **PCI DSS 3.4.1** standards might consider disk encryption sufficient for compliance if you meet these requirements:
+
+   * Separate the logical data access from the operating system authentication.
+
+   * Ensure the decryption key is not linked to user accounts.
+
```
janwieremjewicz, line 44: PCI DSS 3.4.1 is considered retired after 31 March 2025. Maybe worth menioning?

documentation/docs/faq.md
```diff
@@ -25,12 +25,14 @@ Using TDE helps you avoid the following risks:
 
 If to translate sensitive data to files stored in your database, these are user data in tables, temporary files, WAL files. TDE has you covered encrypting all these files.
 
+`pg_tde` does not encrypt system catalogs yet. This is planned for future releases. 
+
```
janwieremjewicz, line 29: Let's make sure that we can elaborate on this. As in what leak vector would that lead to. [Systems catalog](https://www.postgresql.org/docs/current/catalogs.html#:~:text=The%20system%20catalogs%20are%20the,system%20catalogs%20are%20regular%20tables.) holds database metadata so the table and column names are what would not be encrypted. Same with indexes. Am I missing anything

documentation/docs/faq.md
janwieremjewicz: Could wa also include the reasoning that since the WAL is cluster-wide we encrypt it as a whole? That the scenarios for usage of WAL later on require the use of a cluster-wide WAL as there is no database specific one hence encryption of the WAL as a whole makes sense.

documentation/docs/faq.md
nastena1606: @dutow could you please comment here?

jeltz: The big issue is statistics, they are stored in the catalog and contain most common values, datagrams and some other less senstive stuff. Other than that, metadata, yes.

documentation/docs/faq.md
dAdAbird: @nastena1606 The sentence “WAL encryption is done globally for the entire database cluster using the global principal key.” might create an impression that WAL is encrypted directly with the global principal key. But WAL uses the same 2-tier key architecture as for the table files - an internal key encrypts WAL, and a global principal key encrypts internal. And my bad, it's not a "global internal key", but rather a "global principal" and WAL internal key. Global here means that the principal key is not for some database but for global objects (like WAL)

documentation/docs/faq.md
nastena1606: Added more details what's not encrypted so far.

documentation/docs/faq.md
nastena1606: Thanks, updated.

PR #406, opened 2025-01-31
PG-1320 Updated considerations in installation

nastena1606:
Added KMIP server to the list of options

Updated formatting for using the count(*) in docs

modified:   documentation/docs/decrypt.md
	modified:   documentation/docs/index.md
	modified:   documentation/docs/install.md
	modified:   documentation/docs/test.md

PG-0

### DIFF
--- a/documentation/docs/index.md
+++ b/documentation/docs/index.md
@@ -2,9 +2,7 @@
 
 `pg_tde` is the open source PostgreSQL extension that provides Transparent Data Encryption (TDE) to protect data at rest. This ensures that the data stored on disk is encrypted, and no one can read it without the proper encryption keys, even if they gain access to the physical storage media. 
 
-You can configure encryption differently for each database, encrypting specific tables in some databases with different encryption keys while keeping others unencrypted.
-
-Lear more [what is Transparent Data Encryption](tde.md#how-does-it-work) and [why you need it](tde.md#why-do-you-need-tde).
+Learn more [what is Transparent Data Encryption](tde.md#how-does-it-work) and [why you need it](tde.md#why-do-you-need-tde).
 
 !!! important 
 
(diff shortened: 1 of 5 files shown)


documentation/docs/index.md
```diff
@@ -50,7 +48,7 @@ The `pg_tde` extension comes in two distinct versions with specific access metho
 
 ### Which version to chose?
 
-The answer is pretty straightforward: for data sets where indexing is not mandatory or index encryption is not required, use the community version and the `tde_heap_basic` access method. Check the [upstream documentation :octicons-link-external-16:](https://github.com/percona/pg_tde/blob/main/README.md) how to get started.
+Use the community version and the `tde_heap_basic` access method for data sets where indexing is not mandatory or index encryption is not required. Check the [upstream documentation :octicons-link-external-16:](https://github.com/percona/pg_tde/blob/main/README.md) how to get started.
```
dutow, line 51: I think we can modify this to don't use the basic, as we are deprecating it.

documentation/docs/install.md
```diff
@@ -4,13 +4,27 @@
 
 You can use the following options to manage encryption keys:
 
-* Use the HashiCorp Vault server. This is the recommended approach. The Vault server configuration is out of scope of this document. We assume that you have the Vault server up and running. For the  `pg_tde` configuration, you need the following information:
+* Use the Key Management Store (KMS). This is the recommended approach. `pg_tde` supports the following KMS:
 
-    * The secret access token to the Vault server
-    * The URL to access the Vault server
-    * (Optional) The CA file used for SSL verification
+    * KMIP server. `pg_tde` has been tested with PyKMIP
```
dutow, line 9: I wouldn't mention pykmip as a supported server, that's a dummy server intendede for development. We can say that any kmip server should work, if some doesn't, we have to fix that.

documentation/docs/index.md
nastena1606: Sure, I'll address this in a separate PR. Otherwise after the merge we will already reveal this info ahead of time.

documentation/docs/install.md
nastena1606: Updated

dutow (approved):

PR #421, opened 2025-09-30
PG-1879 Various minor cleanup after repository split

jeltz:
This PR does the following:

- Takes copyright, `.gitignore` and code format config files from `percona/postgres`
- Cleans up those files a bit
- Cleans up code owners
- Removes `meson.build`


.dir-locals.el
```diff
@@ -0,0 +1,19 @@
+;; see also src/tools/editors/emacs.samples for more complete settings
```
dutow, line 1: anybody uses emacs?

COPYRIGHT
```diff
@@ -0,0 +1,52 @@
+Percona Server for PostgreSQL - License
```
dutow: I don't think we want to call this repo Postgres, this is just an extension now

.dir-locals.el
jeltz: No idea, I just copied all editor config files without any thought of which we need and not.

COPYRIGHT
jeltz: That is a good point. What license should it be?

dutow: I meant simply to rename to `pg_tde`. Otherwise the portion parts is true, as we copied some tools.

jeltz: I see. Uodated!

.dir-locals.el
jeltz: I think we can merge it and just delete it in the future.

dutow (approved):

PR #422, opened 2025-09-30
PG-1879 Fix build after repository split

jeltz:
After we split `pg_tde` out from the PostgreSQL repository we need to make sure that it builds again.

We also need to copy the patched frotnend tools so we can build them as separate executables to fit a world where pg_tde is a separate extension and not jsut part of a patched PostgreSQL.

Documentation and readme are not updated yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,11 @@ __pycache__
 /src/bin/pg_tde_archive_decrypt
 /src/bin/pg_tde_change_key_provider
 /src/bin/pg_tde_restore_encrypt
+/fetools/*/pg_tde_basebackup
+/fetools/*/pg_tde_checksums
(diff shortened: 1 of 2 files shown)


tools/copy_fetools.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
dutow, line 1: Is this tool still useful? Now it would just overwrite everything.

AndersAstrand: I think it's useful for when we want to add the tools from other postgres versions.

jeltz: Yeah, I have planned to use it for the other PostgreSQL versions.

PR #423, opened 2025-10-02
PG-1879 PostgreSQL 18 support

jeltz:
This adds support for PostgreSQL 18 to `pg_tde`. This required quite extensive changes to `pg_basebackup` due to it being heavily refactored. Additionally we had to do an ugly fix due to the AIO changes which are relevant in with `io_method = sync`. This code only works with sync IO.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -145,9 +145,15 @@ jobs:
         with:
           name: log-test-global-tde-${{ inputs.pg_version }}-${{ inputs.os }}-${{ inputs.compiler }}-${{ inputs.build_type }}
           path: |
-            src/src/test/*/log
-            src/src/test/*/regression.diffs
-            src/src/test/*/regression.out
-            src/src/test/*/results
-            src/src/test/*/tmp_check
+            postgres/contrib/*/log
+            postgres/contrib/*/regression.diffs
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -196,7 +196,7 @@ libtdexlog.a: $(TDE_XLOG_OBJS)
 # Fetches typedefs list for PostgreSQL core and merges it with typedefs defined in this project.
 # https://wiki.postgresql.org/wiki/Running_pgindent_on_non-core_code_or_development_code
 update-typedefs:
-	wget -q -O - "https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_17_STABLE" | cat - typedefs.list | sort | uniq > typedefs-full.list
+	(wget -q -O - "https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_17_STABLE"; wget -q -O - "https://buildfarm.postgresql.org/cgi-bin/typedefs.pl?branch=REL_18_STABLE") | cat - typedefs.list | sort -u > typedefs-full.list
```
dutow, line 199: this seems to work for now, but it can't cause problems when we format earlier version source code with latest pgindent + typedefs?

jeltz: Yeah, this was a pretty ugly fix to get it working.

PR #425, opened 2025-10-02
PG-1879 Fix CI workflows after repo split

AndersAstrand:
This PR fixes the CI scipts after the repo split.

There are two workflows left untouched because they need further configuration of this repo. (Stormweaver requires the runner to be setup, and codecov requires the bot thingy)

This is just a first stab at making everything else work, there are plenty of improvements to be made.

### DIFF
--- a/.github/workflows/pgindent.yml
+++ b/.github/workflows/pgindent.yml
@@ -2,7 +2,7 @@ name: Format
 on:
   pull_request:
     paths-ignore:
-      - contrib/pg_tde/documentation/**
+      - documentation/**
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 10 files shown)


dutow (approved):
I would just disable the coverage and codechecker tests, as we still need them, we will have to readd these files soon. Otherwise looks good to me.

(I would also keep the meson files, as we want to re-add the meson build to the repo, but we deleted the meson build files anyway)

.github/workflows/pgindent.yml
```diff
@@ -13,19 +13,30 @@ jobs:
       - name: Clone repository
         uses: actions/checkout@v4
         with:
+          path: 'src'
```
jeltz, line 16: WOuldn't it make more sense to call it `pg_tde` everywhere to match `postgres`? I know it was `src` in the old matrix tasks but I am not sure that is a good name.

ci_scripts/build-and-install-psp.sh
```diff
@@ -45,5 +34,6 @@ case "$1" in
         ;;
 esac
 
-meson setup build --prefix "$INSTALL_DIR" --buildtype="$BUILD_TYPE" -Dcassert=true -Dtap_tests=enabled $ENABLE_COVERAGE
-cd build && ninja && ninja install
+
```
jeltz: Seems like an empty line snuck in here.

jeltz (approved):

jeltz:
I personally do not care either way about disable vs remove. It is trivial to revert a commit.

PR #426, opened 2025-10-09
Small make improvements

dutow:
* Allow using any pg_config instead of the one available in the path
* Ignore the dirty status of the kmip submodule so that it doesn't show up in git status with in source make builds

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,7 +139,7 @@ EXTRA_CLEAN = \
 	libtde.a \
 	libtdexlog.a
 
-PG_CONFIG = pg_config
+PG_CONFIG ?= pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 PG_CPPFLAGS = -Isrc/include -Isrc/libkmip/libkmip/include -I$(FETOOLS)/include -I$(includedir)
 include $(PGXS)


Makefile
```diff
@@ -139,7 +139,7 @@ EXTRA_CLEAN = \
 	libtde.a \
 	libtdexlog.a
 
-PG_CONFIG = pg_config
+PG_CONFIG ?= pg_config
```
AndersAstrand, line 142: The normal way of doing this with make is to add PG_CONFIG= on the command line rather than as an environment variable. Why do you prefer to use an environment variable?

Ie `make PG_CONFIG=/foo/bar/bin/pg_config` works as expected without this change. no path editing needed.

dutow: I can set an environment variable once, and it's always there. This way it is also consistent with our pgsm makefile.

jeltz: I am fine either way but the `PG_CONFIG = pg_config` lines comes from PostgreSQL and is in other extensions like [https://github.com/pgaudit/pgaudit/blob/main/Makefile](pgaudit) and [https://github.com/postgres-plr/plr/blob/master/Makefile](plr). But that does not say we cannot imrpove on it.

jeltz:
The commit message is a bit too much markdown compared to what we usually do and I wonder if the markdown isn't also broken. Other than that it looks good.

PR #429, opened 2025-10-09
Enable `-Werror` when building in CI

jeltz:
We want to prevent a ton of warnings to enter our build while at the same time not make life hard for external contributors who run some veyr new version of gcc or clang where there is some new warning.

### DIFF
--- a/ci_scripts/make-build.sh
+++ b/ci_scripts/make-build.sh
@@ -4,6 +4,7 @@ set -e
 
 SCRIPT_DIR="$(cd -- "$(dirname "$0")" >/dev/null 2>&1; pwd -P)"
 PG_CONFIG="$SCRIPT_DIR/../../pginst/bin/pg_config"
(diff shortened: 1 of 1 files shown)


ci_scripts/make-build.sh
```diff
@@ -2,37 +2,25 @@
 
 set -e
 
-ARGS=
-
-for arg in "$@"
```
dutow, line 7: Shouldn't we keep the coverage report support, or at least create a jira ticket or something about implementing it again?

jeltz: We for sure should. I just removed the dead code here which made it seem like we had coverage when we had not. But, yes, we should create a Jira ticket for it.

jeltz: https://perconadev.atlassian.net/browse/PG-1949

PR #431, opened 2025-10-14
PG-1428: Relax existing key check on new principal key creation

dutow:
Previously first we checked that the lookup returns success, and then verified that we didn't get any result. This sounds good in theory, but in practice Akeyless reports an error on empty result sets for the locate operation.

This commit removes the return code check so that Akeyless works properly. KeyringGetKey returns NULL on every error return  path, so this definitely doesn't change any behavior with other providers.

### DIFF
--- a/src/catalog/tde_principal_key.c
+++ b/src/catalog/tde_principal_key.c
@@ -524,11 +524,6 @@ pg_tde_create_principal_key_internal(Oid providerOid,
 
 	key_info = KeyringGetKey(provider, key_name, &return_code);
 
-	if (return_code != KEYRING_CODE_SUCCESS)
-		ereport(ERROR,
-				errmsg("failed to retrieve principal key \"%s\" from key provider \"%s\"", key_name, provider_name),
-				errdetail("%s", KeyringErrorCodeToString(return_code)));
-
 	if (key_info != NULL)
 		ereport(ERROR,
 				errcode(ERRCODE_INVALID_PARAMETER_VALUE),


src/catalog/tde_principal_key.c
```diff
@@ -524,11 +524,6 @@ pg_tde_create_principal_key_internal(Oid providerOid,
 
 	key_info = KeyringGetKey(provider, key_name, &return_code);
 
-	if (return_code != KEYRING_CODE_SUCCESS)
```
jeltz, line 527: I am not entirely convinced that this is the right place to change the code. Maybe it should be handled in the KMIP code, by translating the error into a nice not found, rather than handling this in the generic code. Just because KMIP in practice does not make a difference between error and not found does not mean that other protocols like Vault should not. Or does it make sense to always try to set the key on error even if you for example just hit the wrong server/URL?

jeltz (changes requested):

src/catalog/tde_principal_key.c
dutow: KMIP can make a difference between these. The protocol can return an error, or it can return a success with 0 results. Akeyless, a specific implementation, chose to return an error in this case.

If I suppress this in the kmip `keyringGetKey`, that means I'm suppressing it for every codepath that uses that function.  I choose this place because this seems to be the less wrong: even if there's a failure, `KeyringGenerateNewKeyAndStore` will still fail. Even the "is there already a key?" check is a bit redundant, because we can't guarantee that another process doesn't create the same key between the two calls. I think we added this already exists precheck because that's unlikely, and this check improves the user experience with a better error message.

I would argue that the error message I've deleted was actually misleading: when I first saw that Akeyless fails with this message during key creation, I didn't understand - we are trying to create a key, why do we output an error about retrieving it? At first I thought that first we create the key, and then we try to load it from the server to make sure we persisted the correct key, but I checked and the key wasn't there on the server yet. "Failed to retrieve <key-name-we-are-trying-to-create>" suggests that we expect it to be there.

jeltz: A question is if KMIP and Vault servers when writing refuse to overwrite existing secrets. In that case I think we should not even call read before the write at all but just try a write and see what happens. And then we can make the file provider implement the same behavior for writes.

dutow: That's true for KMIP, for vault it depends on permissions, if the access token has update permissions, the v2 engine will create a new version with the new data.

jeltz: Maybe the logic for if it should be just write or read + write should be moved to the provider type code?

jeltz: > Write another version, the previous version will still be accessible. The -cas flag can optionally be passed to perform a check-and-set operation. If not set the write will be allowed. In order for a write to be successful, cas must be set to the current version of the secret. If set to 0 a write will only be allowed if the key doesn’t exist as unset keys do not have any version information. Also remember that soft deletes do not remove any underlying version data from storage. In order to write to a soft deleted key, the cas parameter must match the key's current version.

https://openbao.org/docs/secrets/kv/kv-v2/

Did not look at the protocol but seems very likely that Vault supports write without update.

jeltz: Yes, it does. https://openbao.org/api-docs/secret/kv/kv-v2/#createupdate-secret

dutow: That describes what I mentioned:

> This endpoint creates a new version of a secret at the specified location. If the value does not yet exist, the calling token must have an ACL policy granting the create capability. If the value already exists, the calling token must have an ACL policy granting the update capability.

It doesn't update the existing system, it creates a new version, as kv-2 is versioned.

jeltz: Yeah, but if you set the `cas` parameter it should cause an error if it already exists.

PR #432, opened 2025-10-14
PG-1879 Use new extension point to register custom AIO callback

jeltz:
Instead of patching the md.c callback's and adding our own we use something more like a preoper extension point which we have now added to Percona's fork of PostgreSQL. This new extension point allows registering a new type of AIO callback.

Now we have this new callback we can make sure to aonly register it for encrypted tables to slightly minimze the overhead of normal tables.

### DIFF
--- a/src/smgr/pg_tde_smgr.c
+++ b/src/smgr/pg_tde_smgr.c
@@ -506,9 +506,10 @@ tde_mdstartreadv(PgAioHandle *ioh,
 				 SMgrRelation reln, ForkNumber forknum, BlockNumber blocknum,
 				 void **buffers, BlockNumber nblocks)
 {
-	/* Load key early: later we are in a critical section */
+	/* Uses a global variable to pass the relation key and IV */
 	aio_tdereln = (TDESMgrRelation *) reln;
 
+	/* Load key early: later we are in a critical section */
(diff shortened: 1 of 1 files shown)


src/smgr/pg_tde_smgr.c
```diff
@@ -439,20 +439,18 @@ static TDESMgrRelation *aio_tdereln = NULL;
  * AIO completion callback for tde_mdstartreadv().
  */
 static PgAioResult
-tde_md_readv_complete(PgAioHandle *ioh, PgAioResult prior_result, uint8 cb_data)
+tde_readv_complete(PgAioHandle *ioh, PgAioResult prior_result, uint8 cb_data)
 {
-	PgAioResult mdres = md_readv_complete(ioh, prior_result, cb_data);
+	PgAioResult result = prior_result;
```
dutow: do we even need this varialbe? we never change it

dutow (approved):
One minor comment.

Another question is, shouldn't we close the old PRs in the postgres repo and go back to using proper PRs there?

src/smgr/pg_tde_smgr.c
jeltz: Good question, will take a look at the code.

PR #433, opened 2025-10-14
PG-1949 Re-implement code coverage

jeltz:
When splitting the repository we broke the code coverage so here we re-implement it based on a mix of the old code and `sanitize.yml`.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -65,7 +65,6 @@ jobs:
           name: sanitizers-testlog
           path: |
             sanitize.log.*
-            src/log
             src/regression_install
             src/regression_install.log
             src/regression.diffs


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,82 @@
...
+    paths-ignore:
+      - documentation/**
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - documentation/**
+
+env:
+  pg_version: 18
```
dutow, line 13: Shouldn't we support coverage for the union of all configurations, so ifdefs / conditionally included files are handled correctly? No idea if codecov supports that or not.

jeltz: We probably should but I want to just get this working. But the commit message should explain this.

jeltz: Also I think running with just one version is much better than not running at all and likely good enoough for some time forward.

jeltz: Updated the commit message.

dutow (approved):

PR #442, opened 2025-10-28
PG-1959: Add support for Vault/OpenBao namespaces 

dutow:
This implements support for the "namespace" feature in Vault Enterprise
and OpenBao, available both on the CLI and on the HTTP interface using
the `X-Vault-Namespace` header.

* Change CI to use OpenBao since in Vault this feature is only available
  in the enterprise version
* Add overloads to the SQL interface which accepts an additional
  namespace parameter.
* If namespace parameter is given, correctly forward this to the Vault
  server

### DIFF
--- a/documentation/docs/global-key-provider-configuration/openbao.md
+++ b/documentation/docs/global-key-provider-configuration/openbao.md
@@ -24,7 +24,7 @@ SELECT pg_tde_add_global_key_provider_vault_v2(
 
 * `provider-name` is the name to identify this key provider
 * `secret_token_path` is a path to the file that contains an access token with read and write access to the above mount point
-* `url` is the URL of the Vault server
+* `url` is the URL of the OpenBao server
 * `mount` is the mount point where the keyring should store the keys
 * [optional] `ca_path` is the path of the CA file used for SSL verification
 * [optional] `namespace` is the namespace within the OpenBao server. Read more about the [namespace support in OpenBao](https://openbao.org/blog/namespaces-announcement/)` (NOTE: You can use a namespace without ca_path. In this case, specify NULL as the ca_path value)


ci_scripts/setup-keyring-servers.sh
```diff
@@ -19,17 +19,22 @@ rm -f /tmp/pykmip.db
 pykmip-server -f "$SCRIPT_DIR/../pykmip-server.conf" -l /tmp/kmip-server.log &
 
 CLUSTER_INFO=$(mktemp)
-vault server -dev -dev-tls -dev-cluster-json="$CLUSTER_INFO" > /dev/null &
+bao server -dev -dev-tls -dev-cluster-json="$CLUSTER_INFO" > /dev/null &
```
jeltz, line 22: I would rather have this change as its own commit.

documentation/docs/global-key-provider-configuration/kmip-openbao.md
```diff
@@ -26,6 +27,7 @@ SELECT pg_tde_add_global_key_provider_vault_v2(
 * `url` is the URL of the Vault server
 * `mount` is the mount point where the keyring should store the keys
 * [optional] `ca_path` is the path of the CA file used for SSL verification
+* [optional] `namespace` is the namespace when using the [namespace support in OpenBao](https://openbao.org/blog/namespaces-announcement/)` (NOTE: use "" for empty ca_path if ca_path is not required)
```
jeltz: Can't you use `NULL` for it?

pg_tde--2.0--2.1.sql
```diff
@@ -0,0 +1,70 @@
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION pg_tde UPDATE TO '2.1'" to load this file. \quit
+
+CREATE FUNCTION pg_tde_add_database_key_provider_vault_v2(provider_name TEXT,
+                                                vault_url TEXT,
+                                                vault_mount_path TEXT,
+                                                vault_token_path TEXT,
+                                                vault_ca_path TEXT,
+                                                vault_namespace TEXT)
```
jeltz, line 9: These should be aligned if we keep doing our horrible code style.

pg_tde--2.0--2.1.sql
```diff
@@ -0,0 +1,70 @@
...
+                                                        vault_namespace TEXT)
+RETURNS VOID
+LANGUAGE SQL
+BEGIN ATOMIC
+    SELECT pg_tde_add_global_key_provider('vault-v2', provider_name,
+                            json_object('url' VALUE vault_url,
+                            'mountPath' VALUE vault_mount_path,
+                            'tokenPath' VALUE vault_token_path,
+                            'caPath' VALUE vault_ca_path,
+                            'namespace' VALUE vault_namespace));
```
jeltz, line 35: Same with these and alignment.

sql/vault_v2_test.sql
```diff
@@ -44,4 +44,26 @@ SELECT pg_tde_change_database_key_provider_vault_v2('vault-v2', 'https://127.0.0
 -- HTTP against HTTPS server fails
 SELECT pg_tde_change_database_key_provider_vault_v2('vault-v2', 'http://127.0.0.1:8200', 'secret', :'root_token_file', NULL);
 
+
+-- Namespaces are supported
```
jeltz: Maybe we should explain a bit better what the test tests. From what I understand it tests that you can add a new key in a different namespace and it does not cause an error.

pg_tde--2.0--2.1.sql
dutow: these already aren't aligned, or aligned the same way as I added the new line.

jeltz: True, some functions are aligned others are not. :cry:

documentation/docs/global-key-provider-configuration/openbao.md
```diff
@@ -26,6 +27,7 @@ SELECT pg_tde_add_global_key_provider_vault_v2(
 * `url` is the URL of the Vault server
 * `mount` is the mount point where the keyring should store the keys
 * [optional] `ca_path` is the path of the CA file used for SSL verification
+* [optional] `namespace` is the namespace when using the [namespace support in OpenBao](https://openbao.org/blog/namespaces-announcement/)` (NOTE: use `NULL` for empty ca_path if ca_path is not required)
```
nastena1606:
```suggestion
* [optional] `namespace` is the namespace within the OpenBao server. Read more about the [namespace support in OpenBao](https://openbao.org/blog/namespaces-announcement/)` (NOTE: use `NULL` for empty ca_path if ca_path is not required)
```

nastena1606: Maybe it makes sense to move the NOTE next to the `ca_path` description?

documentation/docs/global-key-provider-configuration/openbao.md
```diff
@@ -26,6 +27,7 @@ SELECT pg_tde_add_global_key_provider_vault_v2(
 * `url` is the URL of the Vault server
```
nastena1606:
```suggestion
* `url` is the URL of the OpenBao server
```

documentation/docs/global-key-provider-configuration/openbao.md
dutow: That's for the situation "I want to specify the namespace, but I don't need a ca_path, what do I have to write there?".

nastena1606: Maybe then something like "You can use a namespace without `ca_path`. In this case, specify `NULL` as the `ca_path` value" ?

dutow: done

PR #443, opened 2025-10-28
PG-1971 Move TdeCluster.pm tests into this repo

jeltz:
Attempt at reducing the diff in our modified version of PostgreSQL and at the same time being able to more closely follow when some behavior changes in pg_tde making it less work to match pg_tde commit with the right PSP version.

### DIFF
--- a/ci_scripts/perl/PostgreSQL/Test/TdeCluster.pm
+++ b/ci_scripts/perl/PostgreSQL/Test/TdeCluster.pm
@@ -5,6 +5,7 @@ use parent 'PostgreSQL::Test::Cluster';
 use strict;
 use warnings FATAL => 'all';
 
+use Cwd 'abs_path';
 use List::Util                      ();
 use PostgreSQL::Test::RecursiveCopy ();
 use PostgreSQL::Test::Utils         ();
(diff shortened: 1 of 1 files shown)


ci_scripts/perl/PostgreSQL/Test/TdeCluster.pm
```diff
@@ -0,0 +1,339 @@
...
+sub _tde_generate_pg_tde_dir
+{
+	my ($self, $extra) = @_;
+	my $temp_dir = PostgreSQL::Test::Utils::tempdir();
+
+	PostgreSQL::Test::Utils::system_log(
+		'initdb',
+		'-D' => $temp_dir,
+		'--set' => 'shared_preload_libraries=pg_tde',
+		'--set' => 'io_method=sync',
```
AndersAstrand: Will this work with pg17 or do we need to add if majorversion et.c. here?

jeltz: Yes, I had planned to ad an if statement but it worked even without it in the CI. I expected it to fail but apparently there is no error on unknown config params. I should research it further but I tried with a total nonsense one.

jeltz: If true that PostgreSQL ignores typos in params I will be a a bit sad.

jeltz: Fixed.

jeltz:
Fixed typos plus changed from "now" to "instead".

PR #445, opened 2025-10-28
Run sanitizers against PostgreSQL 18 instead of 17

jeltz:
Maybe we want to do a full matrix but for now let's at least run against the latest PostgreSQL version.


jeltz:
Yeah, the reason this is a draft is because I wanted to run the sanitizers and see what failed. Not sure if I will continue this work right now so @artemgavrilov if you want to continue this work feel free to do so.

PR #448, opened 2025-10-30
PG-1946 - Create TDE impact topic

Andriciuc:
moved from old repo

### DIFF
--- a/documentation/docs/index/what-tde-impacts.md
+++ b/documentation/docs/index/what-tde-impacts.md
@@ -6,7 +6,7 @@ This page summarizes how `pg_tde` interacts with core PostgreSQL operations.
 |------|-----------|-----------------|
 | **Backups** | ❌ Yes | Encrypted backups require `pg_tde`-aware tools. See [Backup with WAL encryption](../how-to/backup-wal-enabled.md). |
 | **Restore** | ❌ Yes | Restored data remains encrypted; `pg_tde` transparently handles decryption. See [Restore encrypted backups](../how-to/restore-backups.md). |
-| **Streaming replication** | ⚠️ Partial | Replication requires `pg_tde` on replicas. |
+| **Streaming replication** | ⚠️ Partial | Replicas must load the `pg_tde` shared library (shared_preload_libraries = 'pg_tde'). Without it, the replica will fail on operations involving `tde_heap`. |
 | **Logical replication** | ✅ No | Not affected. |
 | **Monitoring and statistics** | ✅ No | Not affected, including `pg_stat_monitor`. |
 | **Performance** | ⚠️ Slight | Minor CPU overhead due to enabling encryption, most noticeable on random write workloads. |


documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
+# Impact of pg_tde on database operations
+
+This page summarizes how `pg_tde` interacts with core PostgreSQL operations.
+
+| Area | Affected | Notes / Actions |
+|------|-----------|-----------------|
+| **Backups** | ✅ Yes | Encrypted backups require `pg_tde`-aware tools. See [Backup with WAL encryption](../how-to/backup-wal-enabled.md). |
+| **Restore** | ✅ Yes | Restored data remains encrypted; `pg_tde` transparently handles decryption. See [Restore encrypted backups](../how-to/restore-backups.md). |
+| **Streaming replication** | ⚠️ Partial | Replication requires `pg_tde` on replicas. If keys are local, they must be manually copied; shared KMS works automatically. |
```
jeltz: Should we really talk about copying the file keys here? Isn't the real summary that both servers need to be able to access the keys/KMS?

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
...
+| **Logical replication** | ? | ? |
+| **Monitoring and statistics** | ? | how will `pg_stat_monitor` be affected? |
```
jeltz: Not at all.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
+# Impact of pg_tde on database operations
+
+This page summarizes how `pg_tde` interacts with core PostgreSQL operations.
+
+| Area | Affected | Notes / Actions |
+|------|-----------|-----------------|
+| **Backups** | ✅ Yes | Encrypted backups require `pg_tde`-aware tools. See [Backup with WAL encryption](../how-to/backup-wal-enabled.md). |
+| **Restore** | ✅ Yes | Restored data remains encrypted; `pg_tde` transparently handles decryption. See [Restore encrypted backups](../how-to/restore-backups.md). |
+| **Streaming replication** | ⚠️ Partial | Replication requires `pg_tde` on replicas. If keys are local, they must be manually copied; shared KMS works automatically. |
+| **Logical replication** | ? | ? |
```
jeltz: Not affected at all other than `pg_createsubscriber`.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
...
+| **Monitoring and statistics** | ? | how will `pg_stat_monitor` be affected? |
+| **Extensions** | ? | ? |
```
jeltz: Yes, they can be affected.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
...
+| **Maintenance operations** | ? | ? |
```
jeltz: Quite vague.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
...
+| **Upgrades and migrations** | ? | ? |
```
jeltz: Two separate unrelated things.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,16 @@
...
+| **Performance** | ⚠️ Slight | Minor CPU overhead due to enabling encryption. |
```
jeltz: Should we mention that it happens primarily on random writes?

documentation/mkdocs.yml
```diff
@@ -180,7 +181,7 @@ nav:
       - "Fortanix configuration": global-key-provider-configuration/kmip-fortanix.md  
       - "Vault configuration": global-key-provider-configuration/vault.md
       - "Thales configuration": global-key-provider-configuration/kmip-thales.md
-      - "Using OpenBao as a key provider": global-key-provider-configuration/openbao.md
+      - "Using OpenBao as a key provider": global-key-provider-configuration/kmip-openbao.md
```
jeltz: Seems like you accidentally changed this line.

documentation/docs/index/what-tde-impacts.md
Andriciuc: We must remember to reiterate information as much as possible to the users, if it is truly unnecessary I will remove it, but if the case is that it repeats, we'd better keep it.

documentation/docs/index/what-tde-impacts.md
Andriciuc: Hmm, I updated to: ⚠️ Partial | pg_createsubscriber is affected. (by which means? a small blurb how it is affected)

documentation/docs/index/what-tde-impacts.md
Andriciuc: Updated to: ❌ No | `pg_stat_monitor` is not affected. |

documentation/docs/index/what-tde-impacts.md
Andriciuc: Should we recommend the users to refer to our experts page/contact us/support? Do we have any recommendations for them to follow?

documentation/docs/index/what-tde-impacts.md
Andriciuc: Hm, true. I can remove it if it is okay with you.

documentation/docs/index/what-tde-impacts.md
Andriciuc: For sure! I have added this info.

documentation/docs/index/what-tde-impacts.md
Andriciuc: Thank you for that! I split them in two separate categories.

documentation/mkdocs.yml
Andriciuc: Thank you! Fixed!

documentation/docs/index/what-tde-impacts.md
jeltz: I think reiterating it here will encourage people to use it in cases they should not.

documentation/docs/index/what-tde-impacts.md
jeltz: Yeah, remove it.

documentation/docs/index/what-tde-impacts.md
jeltz: That is a good question. We need to properly investigate that but presumably the other server needs to access to KMS and have pg_tde isntalled.

documentation/docs/index/what-tde-impacts.md
jeltz: No, we need to document this properly.

documentation/docs/index/what-tde-impacts.md
Andriciuc: Hm this is also true! Then I will just leave: Replication requires `pg_tde` on replicas.

So it is straightforward.

documentation/docs/index/what-tde-impacts.md
Andriciuc: I will then keep it as it is, a partial answer is still an answer! We can update it as investigations are done in the future.

documentation/docs/index/what-tde-impacts.md
Andriciuc: Done, thank you!

documentation/docs/index/what-tde-impacts.md
Andriciuc: Decided to remove this for now.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,13 @@
...
+
+| Area | Affected | Notes / Actions |
+|------|-----------|-----------------|
+| **Backups** | ❌ Yes | Encrypted backups require `pg_tde`-aware tools. See [Backup with WAL encryption](../how-to/backup-wal-enabled.md). |
+| **Restore** | ❌ Yes | Restored data remains encrypted; `pg_tde` transparently handles decryption. See [Restore encrypted backups](../how-to/restore-backups.md). |
+| **Streaming replication** | ⚠️ Partial | Replication requires `pg_tde` on replicas. |
+| **Logical replication** | ✅ No | Not affected. |
+| **Monitoring and statistics** | ✅ No | Not affected, including `pg_stat_monitor`. |
+| **Performance** | ⚠️ Slight | Minor CPU overhead due to enabling encryption, most noticeable on random write workloads. |
+| **Configuration management** | ❌ Yes | `pg_tde.conf` parameters affect key handling and WAL encryption. |
```
jeltz: I think this line should be removed from the table. It does not add anything.

jeltz: Also `pg_tde.conf` is jsut a name we picked. The file can be called anything or the config could be added to `postgresql.conf` directly.

Andriciuc: Ah I see! Thanks for the clarification, I removed this line.

documentation/docs/index/what-tde-impacts.md
```diff
@@ -0,0 +1,12 @@
+# Impact of pg_tde on database operations
+
+This page summarizes how `pg_tde` interacts with core PostgreSQL operations.
+
+| Area | Affected | Notes / Actions |
+|------|-----------|-----------------|
+| **Backups** | ❌ Yes | Encrypted backups require `pg_tde`-aware tools. See [Backup with WAL encryption](../how-to/backup-wal-enabled.md). |
+| **Restore** | ❌ Yes | Restored data remains encrypted; `pg_tde` transparently handles decryption. See [Restore encrypted backups](../how-to/restore-backups.md). |
+| **Streaming replication** | ⚠️ Partial | Replication requires `pg_tde` on replicas. |
```
dAdAbird: I've checked, and a replica without `shared_preload_libraries = 'pg_tde'` would crash trying to replicate DROP, CREATE, ALTER statements on `tde_heap` tables due to unknown rmgr (`FATAL:  resource manager with ID 140 not registered`). And it won't be able to read from such a table (all the `tde_heap ` that were on a primary before syncing up a new replica). 
So all that is needed on replicas is to preload a shared lib `pg_tde`. Should we clarify that? Or is it implicit from the "Replication requires `pg_tde` on replicas." statement?

Andriciuc: Nice find! It is extremely important to tell users exactly what will happen, so I have updated this statement to this:

Replicas must load the `pg_tde` shared library (shared_preload_libraries = 'pg_tde'). Without it, the replica will fail on operations involving `tde_heap`.

dAdAbird (approved):

PR #454, opened 2025-11-01
Do not start keyring servers for PSP tests

jeltz:
Since the PSP test suite only uses key files there is no reason to start OpenBao and PyKMIP.


AndersAstrand (approved):
Maybe also exclude from installation, but this is definitely an improvement regardless!

PR #455, opened 2025-11-03
Remove github CODEOWNER for .github/

AndersAstrand:
No single person is responsible for this, it's the whole team's responsibility to ensure CI works as expected.


jeltz (approved):
Unless @artemgavrilov has any opinion to the contrary.

PR #456, opened 2025-11-03
Remove support for manually triggering CI workflows

jeltz:
~These just add noise to the config for not gain.~

New description:  I could not think of any reason we would want to do this and in either case it should either be an option for all or none.


jeltz:
Then maybe we should instead allow all tasks to be triggered manually?

artemgavrilov:
Mmm, I didn't find any reasons for me to trigger these workflows manually. I guess we can drop them, and revert in case we will find a use case for that.

PR #458, opened 2025-11-06
PG-2003 Add Akeyless KMS topic

Andriciuc:
Akeyless is confirmed to work, now it needs documenting.

Added initial file.


documentation/docs/global-key-provider-configuration/kmip-akeyless.md
```diff
@@ -0,0 +1,9 @@
+# Akeyless KMIP server configuration
+
+`pg_tde` is compatible with the Akeyless CipherTrust Manager via the KMIP protocol. For a full setup guide, see [the Thales documentation](https://thalesdocs.com/ctp/cm/2.19/reference/kmip-ref/index.html?).
```
dutow: why do we link the thales docs for akeyless?

Andriciuc: @dutow The initial draft will be based on the thales doc as a skeleton version to build up content, please ignore it for the moment as it is still in a draft state. 

Thank you!

PR #459, opened 2025-11-06
PG-2001 Stop using OpenSSL functions deprecated in 1.1.0

AndersAstrand:
PostgreSQL 18 now requires OpenSSL 1.1.1 or newer, and all our supported platforms has atleast this version or newer. Since PostgreSQL now requires a newer OpenSSL, these deprecated functions are not available by default.


jeltz (approved):
Looks good but should likely be merged into main since we for now just have the release branch as an alias for main.

PR #460, opened 2025-11-10
PG-2007 Implement AIO support

jeltz:
The main challenge with adding support for AIO was how to make sure the process completing the issued IO has access to the key to decrypt the buffer read from disk. Especially since AIO callbacks are ran in a critical section which makes things like caching keys read from disk hard to do.
    
The solution in this commit is to follow the pattern of `pgaio_ctl->io_handles` and store the decryption keys in fixed size array in shared memory with space reserved for one key per IO handle. This may seem wasteful but it matches what the AIO subsystem does so it is easy to explain and reason about. The backend issuing IO simply sets the decryption key at the same index in the array as the ID of the IO handle.
    
We do not do anything special for temporary tables despite it being  possible to do the decryption in the issuing backend rather than the completing backend. It is not obvious to me which option would be best and this solution is slightly less code.

Additionally we turn on `io_method = worker` in the tests (the default) and enabled the `io_uring` tests.

### DIFF
--- a/ci_scripts/build-and-install-psp.sh
+++ b/ci_scripts/build-and-install-psp.sh
@@ -43,8 +43,16 @@ case "$1" in
 esac
 
 cd "$PSP_DIR"
-./configure --prefix="$INSTALL_DIR" --enable-debug --enable-tap-tests $ARGS
+
+./configure \
+   --prefix="$INSTALL_DIR" \
+   --enable-debug \
+   --enable-tap-tests \
+   --with-liburing \
+   $ARGS
+
 make install-world -j -s
+
 if [ "$INSTALL_INJECTION_POINTS" = 1 ]; then
     # Injection points extension is not built by default
     make install -j -s -C src/test/modules/injection_points
--- a/ci_scripts/ubuntu-deps.sh
+++ b/ci_scripts/ubuntu-deps.sh
@@ -21,6 +21,7 @@ DEPS=(
     libselinux1-dev
     libssl-dev
     libsystemd-dev
+    liburing-dev
     libxml2-dev
     libxml2-utils
     libxslt1-dev


src/smgr/pg_tde_smgr.c
```diff
@@ -433,7 +434,28 @@ tde_mdopen(SMgrRelation reln)
...
+#include "miscadmin.h"
+#include "storage/aio_subsys.h"
+#include "storage/proc.h"
+
+Size
+TDESmgrShmemSize(void)
+{
+	/* TODO */
+	AioShmemSize();
+	return mul_size(MaxBackends + NUM_AUXILIARY_PROCS, mul_size(io_max_concurrency, sizeof(InternalKey)));
```
dutow: wouldn't it be easier to use `io_handle_count`? It should be the same, but it's already calculated.

jeltz: I don't think it is calculated yet when we reach this point of the but I would double check.

jeltz: Tested it and it was indeed not initialized.

src/smgr/pg_tde_smgr.c
```diff
@@ -433,7 +437,44 @@ tde_mdopen(SMgrRelation reln)
...
 #if PG_VERSION_NUM >= 180000
 
-static TDESMgrRelation *aio_tdereln = NULL;
+/*
+ * Area in shared memory which has the capacity to store one InternalKey per
+ * IO handle in pgaio_ctl->io_handles.
+ *
+ * This way we can still have the backend which issues the IO handle fetching,
+ * decrypting and caching relation keys while the process completing the IO
+ * just looks uses the key associated with the handle ID, i.e. the position in
```
dutow: just looks uses ?

src/smgr/pg_tde_smgr.c
```diff
@@ -433,7 +437,44 @@ tde_mdopen(SMgrRelation reln)
 
 #if PG_VERSION_NUM >= 180000
 
-static TDESMgrRelation *aio_tdereln = NULL;
+/*
+ * Area in shared memory which has the capacity to store one InternalKey per
+ * IO handle in pgaio_ctl->io_handles.
```
dutow: this seems a bit confusing, "in pgaio_ctl->io_handles." could also mean that we are storing internal keys there.

src/smgr/pg_tde_smgr.c
jeltz: Seems like I changed my mind of what the sentence should be mid setence. Will fix.

src/smgr/pg_tde_smgr.c
jeltz: Hmm, I see your point. Will see if I can find a better wording.

jeltz: Fixed!

src/smgr/pg_tde_smgr.c
jeltz: Rewrote that part entirely. It was a mess.

PR #464, opened 2025-11-12
Try to exclude fetools directory from Codecov report and update componenets

jeltz:
Including the `feutils` directory in the totals in the coverage report does not make too much sense since we only test the features relevant to `pg_tde`.

It is a bit tricky to test if a config change like this actually works but let's take stab at it.

Also update components to match current realities.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -7,6 +7,7 @@ coverage:
         target: 90%
         threshold: 1%
         base: auto
+        paths: "!fetools/"
 comment:
   layout: "header, diff, components"
 component_management:


.github/codecov.yml
```diff
@@ -7,6 +7,7 @@ coverage:
         target: 90%
         threshold: 1%
         base: auto
+        paths: "!fetools/"
```
artemgavrilov, line 10: There is also dedicated `ignore` option. Since by default we include everything, then maybe `ignore` will look better than `!`.
https://docs.codecov.com/docs/ignoring-paths

Not sure, so up to you.

jeltz: I thought they did different things and that paths did the thing I wanted but without ability to test I was very unsure. I want these files ideally to be included when I browse the coverage tree but not in the total score. I have no idea if that is possible.

PR #465, opened 2025-11-12
PG-1968 Add support for AES 256 encryption

dAdAbird:
A new GUC option is applied to all newly created keys - both principal and internal. 
We had to change the _keys file format to accommodate 32-byte internal keys. So the keys file migration routine was added at server start.

### DIFF
--- a/src/encryption/enc_aes.c
+++ b/src/encryption/enc_aes.c
@@ -63,6 +63,10 @@ AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, int key_len, co
 	Assert(key_len == 16 || key_len == 32);
 	cipher = key_len == 32 ? cipher_ctr_ecb_256 : cipher_ctr_ecb_128;
 
+	/*
+	 * TODO: Currently, only Ecb (WAL) use cached context. This caching was
+	 * done for optimisation. Do we need it anymore?
+	 */
 	if (*ctxPtr == NULL)
 	{
 		Assert(cipher != NULL);


fetools/pg17/pg_basebackup/pg_basebackup.c
```diff
@@ -683,7 +683,7 @@ StartLogStreamer(char *startpos, uint32 timeline, char *sysidentifier,
 			exit(1);
 		}
 		pg_tde_save_server_key(principalKey, false);
-		TDEXLogSmgrInitWrite(true);
+		TDEXLogSmgrInitWrite(true, 16);
```
jeltz: I don't like seeing a random 16 in the source, but will need to think what the right solution is.

src/encryption/enc_aes.c
```diff
@@ -44,20 +48,28 @@ AesInit(void)
 	cipher_cbc = EVP_aes_128_cbc();
```
AndersAstrand: we should add `_128` suffixes to these I think.

src/encryption/enc_tde.c
```diff
@@ -27,9 +27,9 @@ iv_prefix_debug(const char *iv_prefix, char *out_hex)
 #endif
 
 void
-pg_tde_generate_internal_key(InternalKey *int_key)
+pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
 {
-	if (!RAND_bytes(int_key->key, INTERNAL_KEY_LEN))
+	if (!RAND_bytes(int_key->key, key_len))
```
AndersAstrand, line 56: I think we should have the `Assert()` for allowed key lengths here as well.

src/pg_tde_guc.c
```diff
@@ -51,4 +77,17 @@ TdeGucInit(void)
 							 NULL	/* show_hook */
 		);
 
+	DefineCustomEnumVariable("pg_tde.cipher",	/* name */
```
AndersAstrand, line 71: Do we want to support using different ciphers for principal key and relation/wal keys in the future?

src/smgr/pg_tde_smgr.c
```diff
@@ -103,7 +104,7 @@ tde_smgr_create_key(const RelFileLocatorBackend *smgr_rlocator)
 {
 	InternalKey *key = palloc_object(InternalKey);
 
-	pg_tde_generate_internal_key(key);
+	pg_tde_generate_internal_key(key, TdeKeyLength);
```
AndersAstrand: This commit should use 16 here, shouldn't it? We don't support longer internal keys yet.

src/keyring/keyring_api.c
AndersAstrand, line 1: We should probably log a WARNING or NOTICE if the key loaded from the KMS doesn't match the current setting for `TdeKeyLength`, since they obviously want some other cipher than what they're getting.

Maybe also add a `error_hint()` about how to rotate the principal key?

src/pg_tde_guc.c
```diff
@@ -6,11 +6,37 @@
...
 #include "utils/guc.h"
 
+#include "keyring/keyring_api.h"
 #include "pg_tde_guc.h"
 
 bool		AllowInheritGlobalProviders = true;
 bool		EncryptXLog = false;
 bool		EnforceEncryption = false;
+int			Cipher = TDE_CIPHER_AES_128;
+int			TdeKeyLength = KEY_DATA_SIZE_128;
```
AndersAstrand: Why does this have a `Tde` prefix when the others do not?

fetools/pg17/pg_basebackup/pg_basebackup.c
AndersAstrand: These belong in the same commit as where the called function is changed I believe.

src/access/pg_tde_tdemap.c
```diff
@@ -712,15 +724,16 @@ pg_tde_file_header_read(const char *tde_filename, int fd, TDEFileHeader *fheader
...
-	if (*bytes_read != sizeof(TDEFileHeader)
-		|| fheader->file_version != PG_TDE_FILEMAGIC)
+	if (*bytes_read != sizeof(TDEFileHeader))
 	{
 		ereport(FATAL,
 				errcode_for_file_access(),
 				errmsg("TDE map file \"%s\" is corrupted: %m", tde_filename));
 	}
+
+	return fheader->file_version;
```
AndersAstrand: I'm not sure what we gain by returning this value instead of just using the field from the header structure after the call?

src/access/pg_tde_xlog_keys.c
```diff
@@ -55,6 +57,26 @@ typedef struct WalKeyFileEntry
 	/* IV and tag used when encrypting the key itself */
 	unsigned char entry_iv[MAP_ENTRY_IV_SIZE];
 	unsigned char aead_tag[MAP_ENTRY_AEAD_TAG_SIZE];
+} WalKeyFileEntryOld;
+
+typedef struct WalKeyFileEntry
+{
+	uint32		key_len;
+	CipherType	cipher;			/* Cipher type. We support only AES for now. */
```
AndersAstrand: Should we use `uint32` here as we do with the other enum that's written directly to file?

This probably doesn't matter in practice for this enum as it's unlikely to hold more than a handful of values, but I would prefer to get an error if we add so many entries the size of the enum grows rather than the file format randomly changing because of it.

AndersAstrand (changes requested):
The commit messages needs work since they current reveal very little of what the commits actually do.

I also think we need to discuss the technical debt we take on here by pushing generalization of the file format versions to the future. Adding "old" and "new" versions of all the functions works for now ofc, but means that it's something we need to clean up later.

Maybe it's worth taking on that debt now, I'm not sure.

src/access/pg_tde_tdemap.c
```diff
@@ -80,14 +82,36 @@ typedef struct TDEMapEntry
 	/* IV and tag used when encrypting the key itself */
 	unsigned char entry_iv[MAP_ENTRY_IV_SIZE];
 	unsigned char aead_tag[MAP_ENTRY_AEAD_TAG_SIZE];
+} TDEMapEntryOld;
+
+typedef struct TDEMapEntry
+{
+	uint32		key_len;		/* Part of AAD */
```
jeltz: I believe we discussed this but I have sadly forgot. Why are not the key lengths stored in a separate field instead of looking it up based on algorthm? Will we ever support an algorthm which can have diffeent lengths? As far as I remember OpenSSL treats AES-128 and AES-256 as separate algorithms,

fetools/pg17/pg_basebackup/pg_basebackup.c
AndersAstrand: I think it's fine as an intermediate step before the longer keys are supported in the later commits.

dAdAbird: I'll fix it in the upcoming PR(s) when taking care of fetools

src/access/pg_tde_tdemap.c
dAdAbird: I don't remember we've ever discussed that but I see your point. It might make sense, let me think

src/access/pg_tde_tdemap.c
dAdAbird: It seemed like a cleaner interface to me. But looking at it now, you're right, it doesn't worth changes

src/pg_tde_guc.c
dAdAbird: Good question. I don't know. We can name the option as `pg_tde.default_cipher`, as @artemgavrilov proposed in the feedback to the draft PR. But as for me, this name is a bit confusing now

src/keyring/keyring_api.c
dAdAbird: It might also make sense to check that the Principal Key's length matches that of the Internal Key (and issue a WARNING if it doesn't) whenever we decrypt the latter. What do you think? Or is it an overkill?

src/pg_tde_guc.c
AndersAstrand: As discussed elsewhere, I think it's fine to keep the name `pg_tde.cipher` as we don't have a better suggestion.

src/keyring/keyring_api.c
AndersAstrand: Hmm... yes. that makes sense. I wonder if we want WARNING on them or just LOG or INFO? I think we could spam the warning log _a lot_ otherwise.

src/pg_tde_guc.c
dAdAbird: We decided to leave it like it is for now

src/access/pg_tde_tdemap.c
dAdAbird: done

src/access/pg_tde_tdemap.c
```diff
@@ -39,7 +40,8 @@
 }
 #endif
 
-#define PG_TDE_FILEMAGIC			0x03454454	/* version ID value = TDE 03 */
+#define PG_TDE_FILEMAGIC_OLD		0x03454454	/* version ID value = TDE 03 */
```
dutow: Instead of "OLD", maybe we could call these FILEMAGIC_03 / TDEMapEntry03 or something like that? So that we don't end up with "OldOld" or "Old2" the next time we change something.

src/access/pg_tde_tdemap.c
```diff
@@ -80,6 +82,27 @@ typedef struct TDEMapEntry
 	/* IV and tag used when encrypting the key itself */
 	unsigned char entry_iv[MAP_ENTRY_IV_SIZE];
 	unsigned char aead_tag[MAP_ENTRY_AEAD_TAG_SIZE];
+} TDEMapEntryOld;
+
+typedef struct TDEMapEntry
```
dutow, line 57: Similarly we could use a version in this typename?

src/encryption/enc_aes.c
```diff
@@ -32,32 +32,44 @@
...
 
 	if (*ctxPtr == NULL)
 	{
-		Assert(cipher_ctr_ecb != NULL);
+		Assert(cipher != NULL);
 
 		*ctxPtr = EVP_CIPHER_CTX_new();
 
-		if (EVP_CipherInit_ex(*ctxPtr, cipher_ctr_ecb, NULL, key, NULL, 1) == 0)
+		if (EVP_CipherInit_ex(*ctxPtr, cipher, NULL, key, NULL, 1) == 0)
```
dutow, line 76: Not completely related to the PR, but I realized that now we only cache the context for WAL and nothing else.

For everything else, ctx is always created internally every time.

This is just something we should investigate, as the mixed implementation most likely isn't practical.

Maybe we should also have an assert about the cipher size in this function, when we are reusing an older context?

src/include/pg_tde.h
```diff
@@ -9,6 +9,9 @@
 
 #define TDE_TRANCHE_NAME "pg_tde_tranche"
 
+#define PG_TDE_WAL_KEY_FILE_MAGIC 0x024B4557	/* version ID value = WEK 02 */
+#define PG_TDE_FILEMAGIC		  0x04454454	/* version ID value = TDE 04 */
```
dutow: why do we define the old and new magics at different places?

src/include/pg_tde.h
```diff
@@ -9,8 +9,16 @@
 
 #define TDE_TRANCHE_NAME "pg_tde_tranche"
 
+/*
+ * Only numeric version (the most left byte) should be changed when updating
+ * file foramt. Otherwise, it will brake the migration process.
```
dutow: break

src/access/pg_tde_tdemap.c
```diff
@@ -929,52 +898,91 @@ pg_tde_get_smgr_key(RelFileLocator rel)
...
+	Oid			spcOid;			/* Part of AAD */
+	RelFileNumber relNumber;	/* Part of AAD */
+	uint32 type;				/* Part of AAD */
+	uint32		_unused3;		/* Part of AAD */
+
+	uint8		encrypted_key_data[16]; /* 16 is the max key length for this
+										 * version */
+	uint8		key_base_iv[16];
+
+	uint32		_unused1;		/* Will be 1 in existing files entries. */
```
dutow: is there a reason for the strange unused ordering (3-1-4-2)?

dAdAbird: Good catch. Must be a legacy. I don't see how it matters now, so changed it

src/encryption/enc_aes.c
dAdAbird: I'v added an assert and TODO comment regarding context caching

fetools/pg17/pg_basebackup/pg_basebackup.c
AndersAstrand: Why isn't everything included in this PR?

src/pg_tde_guc.c
AndersAstrand: Why was this marked resolved without any change or comment?

src/access/pg_tde_tdemap.c
```diff
@@ -70,7 +70,7 @@ typedef struct TDEMapEntry
 	uint32		type;			/* Part of AAD */
 	uint32		_unused3;		/* Part of AAD */
 
-	uint8		encrypted_key_data[INTERNAL_KEY_LEN];
+	uint8		encrypted_key_data[INTERNAL_KEY_MAX_LEN];
```
AndersAstrand: This will break when reading existing files since you've now changed this from 16 bytes to 32 without dealing with file versions.

I think this commit should only add the support internally without changing any fileformat or such

src/pg_tde_guc.c
dAdAbird: Oopsie, I, for some reason, thought about prefix in `TDE_CIPHER_AES_128` but blanked out on `TdeKeyLength`. Will fix

fetools/pg17/pg_basebackup/pg_basebackup.c
dAdAbird: I didn't want to bloat this PR even more for the reviewers sake. Plus I hoped to merge something before december. But at this stage it probably makes sense to do another commit here. Will do

src/smgr/pg_tde_smgr.c
```diff
@@ -121,7 +122,7 @@ tde_smgr_create_key_redo(const RelFileLocator *rlocator)
 {
 	InternalKey key;
 
-	pg_tde_generate_internal_key(&key);
+	pg_tde_generate_internal_key(&key, KeyLength);
```
dutow, line 125: This is probably safe, but maybe we should have a testcase that:

1. starts with 128 bit keys
2. creates a table / writes some data into it
3. crashes
4. changes the key length to 256 bits
5. performs recovery

fetools/pg18/pg_resetwal/pg_resetwal.c
```diff
@@ -541,7 +541,7 @@ main(int argc, char *argv[])
 	 * because we want to be sure that everything is checked and ready for
 	 * writing at this point.
 	 */
-	TDEXLogSmgrInitWrite(false);
+	TDEXLogSmgrInitWrite(false, 16);
```
dutow: It seems safe to use a fixed 128 bit here and above in basebackup, but a comment mentioning it could help.

src/include/encryption/enc_tde.h
```diff
@@ -5,22 +5,51 @@
...
+{
+	switch (cipher)
+	{
+		case CIPHER_AES_128:
+			return KEY_DATA_SIZE_128;
+		case CIPHER_AES_256:
+			return KEY_DATA_SIZE_256;
+
+		default:
+			Assert(false);
```
dutow: This could be an error instead, so that we don't silently ignore some data corruption in production builds?

src/include/encryption/enc_tde.h
```diff
@@ -5,22 +5,51 @@
...
+#define MAX_KEY_DATA_SIZE KEY_DATA_SIZE_256 /* maximum 256 bit encryption */
+
+typedef enum CipherType
+{
+	CIPHER_AES_128,
+	CIPHER_AES_256,
+} CipherType;
+
+static inline uint32
+pg_tde_cipher_key_lenght(CipherType cipher)
```
dutow: typo, should be length

src/access/pg_tde_tdemap.c
```diff
@@ -876,5 +885,195 @@ pg_tde_get_smgr_key(RelFileLocator rel)
...
+		}
+
+		/* The type check later, when extracting the principal key */
+		if (FILEMAGIC_VERSION(fheader.file_version) == 3)
+			read_map_entry = map_from_disk_entry_v3;
+		else
+			elog(ERROR, "keys migration: unsupported or corrupted version %d of file \"%s\"", FILEMAGIC_VERSION(fheader.file_version), db_map_path);
+
+		/*
+		 * The old file exists and it's not empty, hece a principal key should
```
dutow: hence

src/include/pg_tde.h
```diff
@@ -9,6 +9,17 @@
 
 #define TDE_TRANCHE_NAME "pg_tde_tranche"
 
+/*
+ * Only numeric version (the most left byte) should be changed when updating
+ * file foramt. Otherwise, it will break the migration process.
```
dutow: format

src/encryption/enc_aes.c
```diff
@@ -32,37 +32,55 @@
...
+AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, int key_len, const unsigned char *in, int in_len, unsigned char *out)
 {
 	int			out_len;
+	const EVP_CIPHER *cipher;
 
+	Assert(key_len == 16 || key_len == 32);
+	cipher = key_len == 32 ? cipher_ctr_ecb_256 : cipher_ctr_ecb_128;
+
+	/*
+	 * TDOD: Currently, only Ecb (WAL) use cached context. This caching was
```
dutow: TODO

src/access/pg_tde_tdemap.c
```diff
@@ -876,5 +885,195 @@ pg_tde_get_smgr_key(RelFileLocator rel)
...
+
+	memcpy(out->base_iv, entry->key_base_iv, sizeof(entry->key_base_iv));
+	if (!AesGcmDecrypt(principal_key->keyData, principal_key->keyLength,
+					   entry->entry_iv, sizeof(entry->entry_iv),
+					   (unsigned char *) entry, offsetof(TDEMapEntryV3, encrypted_key_data),
+					   entry->encrypted_key_data, out->key_len,
+					   out->key,
+					   entry->aead_tag, sizeof(entry->aead_tag)))
+		ereport(ERROR,
+				errmsg("Failed to decrypt key, incorrect principal key or corrupted key file"));
```
dutow: shouldn't error messages start with lowercase letters?

src/pg_tde.c
```diff
@@ -76,7 +85,8 @@ tde_shmem_startup(void)
 	TDEXLogSmgrShmemInit();
 
 	TDEXLogSmgrInit();
-	TDEXLogSmgrInitWrite(EncryptXLog);
+	pg_tde_migrate_internal_keys();
```
dutow, line 85: Are we safe if migration happens during recovery? I think so, but maybe this could be another testcase with fixed keyfiles/datadir. Database crashes with previous shared library version, still has old files, migration code starts, runs, ...

dutow:
I wanted to post these only as a comment, but I pressed the wrong button.

fetools/pg18/pg_resetwal/pg_resetwal.c
dAdAbird: I'll deal with FE tools that in the upcoming commit.

dAdAbird: For pg_resetwal, the key size doesn't matter at all as we always generate unencrypted record here (I'll add a comment). As for basebackup, I was thinking of adding a new flag to the set the key size. But now I'm not sure if it worth the effort...

src/encryption/enc_aes.c
```diff
@@ -32,37 +32,55 @@
...
-AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, const unsigned char *in, int in_len, unsigned char *out)
+AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, int key_len, const unsigned char *in, int in_len, unsigned char *out)
 {
 	int			out_len;
+	const EVP_CIPHER *cipher;
 
+	Assert(key_len == 16 || key_len == 32);
+	cipher = key_len == 32 ? cipher_ctr_ecb_256 : cipher_ctr_ecb_128;
+
+	/*
```
AndersAstrand, line 66: I don't think this TODO belongs in this commit as it has nothing to do with adding support for 256 bit encryption

src/access/pg_tde_xlog_keys.c
```diff
@@ -45,7 +45,7 @@ typedef struct WalKeyFileEntry
 	uint32		_unused1;		/* Part of AAD, is 1 or 2 in existing entries */
 	uint32		_unused2;		/* Part of AAD */
 
-	uint8		encrypted_key_data[INTERNAL_KEY_LEN];
+	uint8		encrypted_key_data[INTERNAL_KEY_MAX_LEN];
```
AndersAstrand: same here with breaking existing files

src/smgr/pg_tde_smgr.c
dAdAbird: test added

fetools/pg17/pg_basebackup/pg_basebackup.c
```diff
@@ -683,7 +688,18 @@ StartLogStreamer(char *startpos, uint32 timeline, char *sysidentifier,
 			exit(1);
 		}
 		pg_tde_save_server_key(principalKey, false);
-		TDEXLogSmgrInitWrite(true);
+
+		/*
+		 * If no cipher was specified then we try to get the key length from
+		 * the principal key. And resort to 128 bit if no luck with that.
+		 */
+		if (encrypt_wal_key_len != KEY_DATA_SIZE_DEFAULT &&
```
dutow: Isn't this condition inverted? "If no cipher was specified"

t/crash_recovery.pl
```diff
@@ -112,6 +112,21 @@
 );
 PGTDE::poll_start($node);
 
+PGTDE::psql($node, 'postgres', "INSERT INTO test_enc (x) VALUES (7), (8);");
+PGTDE::append_to_result_file("-- kill -9");
+$node->kill9;
+PGTDE::append_to_result_file("-- change chipher to aes_256");
```
dutow: typo: chipher

src/include/encryption/enc_tde.h
```diff
@@ -5,22 +5,37 @@
 #ifndef ENC_TDE_H
 #define ENC_TDE_H
 
-#define INTERNAL_KEY_LEN 16
+#define TDE_KEY_NAME_LEN 256
+#define KEY_DATA_SIZE_128 16	/* 128 bit encryption */
+#define KEY_DATA_SIZE_256 32	/* 256 bit encryption, not yet supported */
```
dutow: "Not yet supported"?

fetools/pg17/pg_basebackup/pg_basebackup.c
dAdAbird: yes, good catch, thanks

src/pg_tde.c
dAdAbird: Added a TAP test

src/access/pg_tde_tdemap.c
```diff
@@ -406,10 +406,10 @@ pg_tde_initialize_map_entry(TDEMapEntry *map_entry, const TDEPrincipalKey *princ
 				errcode(ERRCODE_INTERNAL_ERROR),
 				errmsg("could not generate iv for key map: %s", ERR_error_string(ERR_get_error(), NULL)));
 
-	AesGcmEncrypt(principal_key->keyData,
+	AesGcmEncrypt(principal_key->keyData, principal_key->keyLength,
 				  map_entry->entry_iv, MAP_ENTRY_IV_SIZE,
 				  (unsigned char *) map_entry, offsetof(TDEMapEntry, encrypted_key_data),
-				  rel_key_data->key, INTERNAL_KEY_LEN,
+				  rel_key_data->key, INTERNAL_KEY_MAX_LEN,
```
AndersAstrand: Doesn't feel right to pass 32 here when we use 16 elsewhere in this commit. I find this confusing.

src/access/pg_tde_tdemap.c
```diff
@@ -585,10 +585,10 @@ tde_decrypt_rel_key(const TDEPrincipalKey *principal_key, TDEMapEntry *map_entry
 
 	Assert(principal_key);
 
-	if (!AesGcmDecrypt(principal_key->keyData,
+	if (!AesGcmDecrypt(principal_key->keyData, principal_key->keyLength,
 					   map_entry->entry_iv, MAP_ENTRY_IV_SIZE,
 					   (unsigned char *) map_entry, offsetof(TDEMapEntry, encrypted_key_data),
-					   map_entry->encrypted_key_data, INTERNAL_KEY_LEN,
+					   map_entry->encrypted_key_data, INTERNAL_KEY_MAX_LEN,
```
AndersAstrand: Same here with 32 when we have 16 elsewhere.

src/encryption/enc_tde.c
```diff
@@ -27,9 +27,11 @@ iv_prefix_debug(const char *iv_prefix, char *out_hex)
 #endif
 
 void
-pg_tde_generate_internal_key(InternalKey *int_key)
+pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
 {
-	if (!RAND_bytes(int_key->key, INTERNAL_KEY_LEN))
+	Assert(key_len == 16 || key_len == 32);
+
+	if (!RAND_bytes(int_key->key, key_len))
```
AndersAstrand, line 56: We should probably explicitly zero the unused last 16 bytes of `int_key->key` when `key_len` is 16, or always generate `INTERNAL_KEY_MAX_LEN` bytes regardless. Doesn't feel right to leave it to chance what those bytes contain.

src/pg_tde_guc.c
AndersAstrand: The question still stands. Why does this have a Tde prefix?

fetools/pg17/pg_basebackup/pg_basebackup.c
```diff
@@ -683,7 +688,18 @@ StartLogStreamer(char *startpos, uint32 timeline, char *sysidentifier,
 			exit(1);
 		}
 		pg_tde_save_server_key(principalKey, false);
-		TDEXLogSmgrInitWrite(16);
+
+		/*
+		 * If no cipher was specified then we try to get the key length from
+		 * the principal key. And keep 128 bit if no luck with that.
```
AndersAstrand, line 694: I think the interaction between this `if ()` clause and the `?:` above is confusing. Maybe add it all here as normal if/else stuff?

fetools/pg17/pg_resetwal/pg_resetwal.c
```diff
@@ -518,8 +518,10 @@ main(int argc, char *argv[])
 	 * We are doing a write initialization only here and not at the startup
 	 * because we want to be sure that everything is checked and ready for
 	 * writing at this point.
+	 *
+	 * Key size does not matter as we write unencrypted WAL here.
 	 */
-	TDEXLogSmgrInitWrite(false);
+	TDEXLogSmgrInitWrite(false, KEY_DATA_SIZE_128);
```
AndersAstrand: Why do we need to call this at all if we write unencrypted wal? I think the comment might need some more explanation or should this be changed to only init the Rmgr and not the "XLogSmgr"?

Is it because we need to be able to _read_ encrypted wal and that parameter is ignored for reading? If so maybe the interface for TDEXLogSmgrInitWrite() should be different?

fetools/pg18/pg_basebackup/pg_basebackup.c
```diff
@@ -683,7 +688,18 @@ StartLogStreamer(char *startpos, uint32 timeline, char *sysidentifier,
 			exit(1);
 		}
 		pg_tde_save_server_key(principalKey, false);
-		TDEXLogSmgrInitWrite(16);
+
+		/*
+		 * If no cipher was specified then we try to get the key length from
+		 * the principal key. And keep 128 bit if no luck with that.
+		 */
+		if (encrypt_wal_key_len == KEY_DATA_SIZE_DEFAULT &&
```
AndersAstrand: same here about consolidating this `if` and the `?:` above

src/access/pg_tde_xlog_keys.c
```diff
@@ -194,12 +216,13 @@ pg_tde_create_wal_range(WalEncryptionRange *range, WalEncryptionRangeType type)
 
 	/* TODO: no need in generating key if WAL_ENCRYPTION_RANGE_UNENCRYPTED */
 	range->type = type;
+
```
AndersAstrand: What's up with this added newline?

src/access/pg_tde_xlog_keys.c
```diff
@@ -615,12 +638,13 @@ pg_tde_wal_range_from_entry(const TDEPrincipalKey *principal_key, WalKeyFileEntr
...
 	range->end.lsn = MaxXLogRecPtr;
+	range->key.key_len = pg_tde_cipher_key_lenght(entry->cipher);
 
 	memcpy(range->key.base_iv, entry->key_base_iv, INTERNAL_KEY_IV_LEN);
-	if (!AesGcmDecrypt(principal_key->keyData, 16,
+	if (!AesGcmDecrypt(principal_key->keyData, principal_key->keyLength,
 					   entry->entry_iv, MAP_ENTRY_IV_SIZE,
-					   (unsigned char *) entry, offsetof(WalKeyFileEntry, encrypted_key_data),
-					   entry->encrypted_key_data, 16,
+					   (unsigned char *) entry, offsetof(WalKeyFileEntry, range_start),
```
AndersAstrand, line 625: I don't think we should use the raw struct data for AAD here, since it's not at all obvious whether this will include alignment holes or not et.c.

I would prefer to include the range_start in the aad. Right now there is only one bit that will differ between entries which is not nearly enough to be useful.

src/access/pg_tde_tdemap.c
```diff
@@ -63,14 +66,14 @@ typedef struct TDEFileHeader
  * encrypting/decrypting existing keys from the key files, so any changes here
  * might break existing clusters.
  */
-typedef struct TDEMapEntry
+typedef struct TDEMapEntryOld
```
AndersAstrand: Since this Old struct is unused maybe it shouldn't be included in this commit at all?

src/access/pg_tde_tdemap.c
```diff
@@ -408,8 +429,8 @@ pg_tde_initialize_map_entry(TDEMapEntry *map_entry, const TDEPrincipalKey *princ
 
 	AesGcmEncrypt(principal_key->keyData, principal_key->keyLength,
 				  map_entry->entry_iv, MAP_ENTRY_IV_SIZE,
-				  (unsigned char *) map_entry, offsetof(TDEMapEntry, encrypted_key_data),
-				  rel_key_data->key, INTERNAL_KEY_MAX_LEN,
+				  (unsigned char *) map_entry, offsetof(TDEMapEntry, entry_iv),
```
AndersAstrand, line 401: Again here we shouldn't use the raw struct data for AAD I think. Or if we do we should make sure we at least don't have alignment holes included in the data by defining unused fields where the holes are.

src/access/pg_tde_tdemap.c
```diff
@@ -908,3 +894,186 @@ pg_tde_get_smgr_key(RelFileLocator rel)
...
+#ifndef FRONTEND
+
+/*****************************************
+ * Functions for migrating old smgr keys into a new format file.
+ *****************************************/
+
+/*
+ * A version-specific migration routine. It reads an entry of the specific
+ * version from the given fd and offset, and transforms it into
+ * WalKeyFileEntry (current version)
```
AndersAstrand: This should be TDEMapEntry rather than WalKeyFileEntry, right?

src/include/pg_tde.h
```diff
@@ -9,6 +9,17 @@
 
 #define TDE_TRANCHE_NAME "pg_tde_tranche"
 
+/*
+ * Only numeric version (the most left byte) should be changed when updating
```
AndersAstrand, line 13: Hmm... I wonder if we shouldn't just use the whole filemagic as version instead of parsing it. If we want to split it we should probably have MAGIC and VERSION separate in the header of the new file versions. so we don't rely on parsing the magic tag in the future.

src/access/pg_tde_tdemap.c
```diff
@@ -908,3 +894,186 @@ pg_tde_get_smgr_key(RelFileLocator rel)
...
+
+		/* check if we have anything to do */
+		if (fheader.file_version == PG_TDE_SMGR_FILE_MAGIC)
+		{
+			CloseTransientFile(old_fd);
+			continue;
+		}
+
+		/* The type check later, when extracting the principal key */
+		if (FILEMAGIC_VERSION(fheader.file_version) == 3)
```
AndersAstrand, line 1042: I think we can ditch the macro here and just do something like `if (fheader.file_version == PG_TDE_FILE_MAGIC_V3)` here and in other similar places.

fetools/pg17/pg_resetwal/pg_resetwal.c
dAdAbird: This functions if initializing WAL writes, not necessary encrypped. TDEXLogSmgrInitWrite basically generates a new WAL range. It is unfortunate that now it needs a key size even when it is not needed. I'll think, maybe we should accept `0` as the second arg. And, yeah, I'll add a comment there. May be we should change the logic around  TDEXLogSmgrInitWrite, but I don't feel like should squeeze it in this PR as well.

src/access/pg_tde_xlog_keys.c
dAdAbird: Yeah, fair enough, I'll take a look.

src/include/pg_tde.h
dAdAbird: Splitting it will cause padding issues for previous versions. char[3] + short -> 6 or 8 bytes instead of 4.

src/access/pg_tde_tdemap.c
dAdAbird: My idea is to avoid all these _V%d constants. I don't like the perspective of having bunch of constants `PG_TDE_FILE_MAGIC_V3`, `PG_TDE_FILE_MAGIC_V4` ... `_V99` which values are all the same but numeric part. May way we just have bump PG_TDE_FILE_MAGIC's version and add transformation/update funcs for the previos version(s) instead of creating new constants.

fetools/pg17/pg_resetwal/pg_resetwal.c
AndersAstrand: Makes sense. I think maybe a 0 that makes it clear that the value don't matter and some assert later that documents that writes are not possible if the value is 0 should be fine.

documentation/docs/how-to/backup-wal-enabled.md
```diff
@@ -9,13 +9,13 @@ Also copy any external files referenced by your providers configuration (such as
...
     ```bash
-    pg_tde_basebackup -D /path/to/backup -E
+    pg_tde_basebackup -D /path/to/backup -E [aes_128|aes_256]
     ```
 
     Where:
 
     - `-D /path/to/backup` specifies the backup location where you have to copy `pg_tde`.
-    - `-E` (or `--encrypt-wal`) enables WAL encryption and validates that the copied `pg_tde` and provider files are present and that the server key is accessible (required).
+    - `-E` (or `--encrypt-wal`) enables WAL encryption and validates that the copied `pg_tde` and provider files are present and that the server key is accessible (required). `-E` optionally accepts a cipher argument. Currently available options are `aes_128` and `aes_256`. If cipher is not set, pg_tde_basebackup will try get it from the server key, and resort to `aes_128` if not available.
```
Andriciuc: The `-E` option accepts a cipher argument. The currently supported values are `aes_128` and `aes_256`. If `-E` is not specified, pg_tde_basebackup attempts to determine the cipher from the server's key. If no cipher can be determined, pg_tde_basebackup defaults to `aes_128`.

documentation/docs/variables.md
```diff
@@ -64,3 +64,12 @@ You can set this variable at the following levels:
...
     Setting this variable doesn't affect existing uses of global keys. It only prevents the creation of new principal keys using global providers.
+
+## pg_tde.cipher
+
+**Type** - string <br>
+**Default** - aes_128
+
+A `string` variable to choose a cipher (encryption algorithm). Currently, only `aes_128` and `aes_256` options are available, corresponding to AES 128-bit and 256-bit key lengths, respectively. 
+
+The new setting will be applied only to newly created objects (Primary keys, internal keys, and data encrypted by those keys). All the existing objects will NOT be re-encrypted.
```
Andriciuc: A `string` variable that selects the cipher (encryption algorithm). Currently, the supported values are `aes_128` and aes_256`, corresponding to AES 128-bit and 256-bit key lengths, respectively.

The setting applies only to objects created after the value is set, including principal keys, internal keys, and data encrypted by those keys. Existing objects are not re-encrypted.

Andriciuc (approved):
LGTM, with minor text improvements for clarity!

src/access/pg_tde_xlog_keys.c
dAdAbird: We can't include `range_start` into AAD as it is not a constant right now. The key is generated and first written with an empty (zero) `range_start` as it gets updated later on with the first WAL write. I'll leave it as it is and add a TODO comment. 

> Right now there is only one bit that will differ between entries which is not nearly enough to be useful.

AFAIK, AAD has nothing to do with the robustness of the encryption. It's just an additional unencrypted data associated with the encryption, so we can verify that it wasn't changed. In that sense, we may want to add `range_start` to the AAD. But it's not that easy right now. I'll leave a TODO for that

As for the padding, alignment, etc., I don't think it's an issue right now - two uint32 won't be padded on 32bit nor 64-bit platforms (it doesn't make sense). Even if so, AAD just verifies that the associated data hasn't changed. And we don't update/rewrite this part once it was written the first time.

AndersAstrand: I think we should just do `NULL, 0` here instead of useless AAD then, but if you want to keep it as is that's fine. We can change it in the next file format version 🙂.

src/include/encryption/enc_tde.h
AndersAstrand: This still needs to be fixed

src/include/encryption/enc_tde.h
AndersAstrand: If not an error, we should probably use the `pg_unreachable()` macro instead.

src/include/pg_tde.h
AndersAstrand: YEah i meant use the whole filemagic as version. not splitting it into separate fields.

src/access/pg_tde_tdemap.c
AndersAstrand: This should be TDEMapEntry rather than WalKeyFileEntry in this comment, right?

src/access/pg_tde_xlog_keys.c
dAdAbird: Unfortunately `NULL, 0` won't work as AesGcmEncrypt/Decrypt doesn't understand this. We probably have to change those funds but I'd prefer to do it in another PR. I'll leave a TODO comment regarding AAD here

dAdAbird: I extended an already existent TODO (where we call AesGcmEncrypt)

PR #467, opened 2025-11-13
PG-1932 Make sure archive and restore comamnds do not leak files

jeltz:
As reported by Roman Leventov in https://github.com/percona/postgres/pull/582 we had an issue where we easily could leak temporary files and directories when on failure, especially when the wrapped restore or archive command fails.

The solution in this commit is to move the cleanup of temporary files and directories to an `atexit()` callback plus add a signal handler for `SIGINT` and `SIGTERM`. The signal handler makes sure to kill itself after running cleanup so the parent process gets the right error.

We have a tiny race condition between creating the temporary directory and registering the signal handler but that does not seem worth fixing since it would just leak a directory, not the WAL file, the race conditions being tiny.

With the current hardcoded path of `/dev/shm` and the TAP test framework testing this seems a pain so no tests are included.

Question: Should I add TAP tests or is it ok that I skipped them?

### DIFF
--- a/src/bin/pg_tde_archive_decrypt.c
+++ b/src/bin/pg_tde_archive_decrypt.c
@@ -8,6 +8,8 @@
 #include "access/pg_tde_fe_init.h"
 #include "access/pg_tde_xlog_smgr.h"
 
+#include <signal.h>
+
 #define TMPFS_DIRECTORY "/dev/shm"
 
 static bool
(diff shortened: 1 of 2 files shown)


dAdAbird (approved):
LGTM overall. The only thing, maybe it worth putting a comment about that race somewhere in the code? So it'd be less questions and investigation when someone looks at the code later on.

PR #471, opened 2025-11-17
Clean up curl code a bit

jeltz:
The curl code was a bit messy so let's clean it up slightly.

Changes:

- Remove irrelevant option for enabling STARTTLS
- Use boolean literals
- Do not set values to their defaults
- Stop using long literals and casting to long.


AndersAstrand (changes requested):
The second commit message should probably also mention what happens if curl_easy_init() is used instead of curl_easy_reset().

PR #469, opened 2025-11-14
PG-2012 - Rename tools with pg_tde prefix

Andriciuc:
Updated all mentions of pg_{component} with pg_tde_{component}

PR #477, opened 2025-11-18
PG-2012 Update Patroni example to use our custom frontend tools

jeltz:
I am not sure if we actually need to run our own `pg_basebackup` but promoting using it makes a ton of sense.

TODO: Investigate if this is actually right for `pg_basebackup`.


jeltz:
Part of #469

PR #478, opened 2025-11-18
Add CI workflow for OSSF Scorecard

artemgavrilov:
Restore worklow that uploaded OSSF scorecard results.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,37 @@
+name: Scorecard
+on:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: "24 3 * * 1"
+  push:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Analysis
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      id-token: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
+          persist-credentials: false
+
+      - name: Run analysis
+        uses: ossf/scorecard-action@f49aabe0b5af0936a0987cfb85d86b75731b0186 # v2.4.1
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          publish_results: true
+
+      - name: Upload to code-scanning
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
jeltz: Maybe we should have a commit where we update this to use the latest version. Plus I think we just write `@v5` in our other actions.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Upload results
+        uses: actions/upload-artifact@97a0fba1372883ab732affbe8f94b823f91727db # v3.pre.node20
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard (optional).
+      - name: "Upload to code-scanning"
```
jeltz: Remove quotes.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: "24 3 * * 1"
+  push:
+    branches:
+      - main
+
+# Declare default permissions as read only.
```
jeltz: Kinda useless comment but is this copied from their code and if so do we want to keep ours the same as the original?

.github/workflows/scorecard.yml
artemgavrilov: There is recommendation from OpenSSF to pin action by hash instead of version to prevent CI attacks. It affects project score. I wanted to let dependabot update deps with proper hashes.

artemgavrilov: Updated manually

.github/workflows/scorecard.yml
artemgavrilov: Done

.github/workflows/scorecard.yml
artemgavrilov: Removed

PR #493, opened 2025-12-05
Add namespace note update in the Vault configuration chapter

Andriciuc:


documentation/docs/global-key-provider-configuration/vault.md
```diff
@@ -60,6 +60,29 @@ vault secrets enable -path=tde -version=2 kv
...
+        NULL
+    );
+    ```
+
+    This is equivalent to configuring a namespace separately and works with both:
+
+    * Vault OSS  
+    * Vault Enterprise (with namespaces)
+
+    Using the full path (`namespace/mount/...`) is the recommended approach.
```
dutow: I wouldn't say that this is the recommended way - after our rerelease fix there should be no difference, and there are some valid reasons why people might want to use proper namespaces.

Andriciuc: Decided to remove this then, thank you for the review!

PR #494, opened 2025-12-10
PG-2143 - Redesign the Architecture chapter

Andriciuc:
This PR focuses on reorganizing the information in the Architecture chapter, improve wording, flow and information given to user to be more easily digestible.

We have a huge architecture chapter, one page. As pg_tde continues to evolve, we'll split this chapter into easily editable chunks of information.

The Technical reference overview now has added cards for a quick skim over the contents of the chapter. The **About pg_tde architecture** is split into specific chapters for easier readability.

The former User interface topic is now it's own chapter called Usage guide, and contains all the information needed to deploy or operate pg_tde in their PostgreSQL environment.

### DIFF
--- a/documentation/docs/architecture/encryption-architecture.md
+++ b/documentation/docs/architecture/encryption-architecture.md
@@ -81,4 +81,4 @@ With these modifications, `pg_tde` implements an additional layer on top of the 
 
 WAL encryption is implemented through a separate, server-wide mechanism that extends PostgreSQL WAL-related APIs. Similar to the SMGR integration, this relies on additional extensions to PostgreSQL core APIs to support encryption of WAL records.
 
-For details, see [Configure WAL encryption](../wal-encryption.md).
+For configuration details, see [Configure WAL encryption](../wal-encryption.md).


documentation/docs/advanced-topics/usage-guide.md
```diff
@@ -0,0 +1,171 @@
...
+To change a provider from a command line, `pg_tde` provides the `pg_tde_change_key_provider` command line tool.
+
+This tool works similarly to the above functions, with the following syntax:
+
+```sh
+pg_tde_change_key_provider <dbOid> <providerType> ... details ...
+```
+
+!!! note
+    Since this tool is expected to be offline, it bypasses all permission checks. This is also the reason why it requires a `dbOid` instead of a name, as it has no way to process the catalog and look up names.
```
dAdAbird: > Since this tool is expected to be offline

Do you mean the database is offline? I don't quite understand this statement

documentation/docs/advanced-topics/tech-reference.md
```diff
@@ -1,16 +1,39 @@
...
+
+### :material-tune: GUC Variables { .title }
+
+Configure how `pg_tde` behaves with PostgreSQL. Control WAL encryption, enforce encryption for new tables, and manage global provider inheritance. Includes scope levels, defaults, and permission requirements.
+
+[Configure GUC variables :material-arrow-right:](../variables.md){.md-button}
+
+</div><div data-banner markdown>
+
+### :material-database-sync: Streaming replication { .title }
```
dAdAbird: It's just a nitpicking, but this icon doesn't look like a replication but rather a disk sync. Maybe some arrows w/o that disk?

documentation/docs/advanced-topics/usage-guide.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Provider permissions
+
+`pg_tde` implements access control based on execution rights on the administration functions.
+
+For keys and providers administration, it provides two pair of functions:
+
+```sql
+pg_tde_GRANT_database_key_management_TO_role
+pg_tde_REVOKE_database_key_management_FROM_role
```
dAdAbird: The rest of the function examples are with parenthesis but these two are not. And they expect an argument, don't they?

documentation/docs/advanced-topics/usage-guide.md
```diff
@@ -0,0 +1,171 @@
...
+CREATE TABLE t1(a INT) USING tde_heap;
+```
+
+## Alter an encrypted table
+
+To alter or modify an encrypted table, use the following command:
+
+```sql
+ALTER TABLE t1 SET ACCESS METHOD tde_heap;
+```
```
dAdAbird, line 167: This example shows how to make an unencrypted table encrypted. It doesn't modify anything if the table is encrypted.  Altering of an encrypted table (create or change an index, add a column etc) doesn't require any extra steps (AFIK)

documentation/docs/architecture/encryption-architecture.md
```diff
@@ -0,0 +1,78 @@
...
+* `AES-128-CTR` for WAL encryption; encrypted with internal keys.
+* `AES-128-GCM` for encrypting internal keys; encrypted with the principal key.
+
+## Encryption workflow
+
+You can use `pg_tde` to encrypt entire databases or only selected tables.
+
+To support this without metadata changes, encrypted tables are labeled with the `tde_heap` access method marker.
+
+The `tde_heap` access method is functionally identical to the `heap` access method. It uses the same functions internally without any changes, but with a different name and ID. This allows `pg_tde` to distinguish between encrypted (`tde_heap`)  and non-encrypted (`heap`) tables.
```
dAdAbird: > It uses the same functions internally without any changes, but with a different name and ID.

I struggle to understand what does it mean. What "different name and ID." and in which functions?

documentation/docs/architecture/encryption-architecture.md
```diff
@@ -0,0 +1,78 @@
...
+
+The initial decision about encryption is made using the `postgres` event trigger mechanism:
+
+* When the `tde_heap` clause is used for `CREATE TABLE` or `ALTER TABLE` statements, then the newly created data files are marked as encrypted
+* After this, the file operations encrypt or decrypt the data
+
+Subsequent decisions are done using a slightly modified Storage Manager (SMGR) API:
+
+* When a database file is re-created with a different ID as a result of a `TRUNCATE` or a `VACUUM FULL` command, the newly created file inherits the encryption information
+* The file is then either encrypted or left unencrypted based on that inheritance
```
dAdAbird, line 46: Is it intentional that unordered lists don't have full stops in the end of point but ordered do?

documentation/docs/architecture/encryption-architecture.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Encrypting other access methods
+
+Currently `pg_tde` only encrypts `heap` tables and other files such as indexes, TOAST tables, sequences that are related to the `heap` tables.
+
+Indexes include any kind of index that goes through the SMGR API, not just the built-in indexes in PostgreSQL.
+
+Other table access methods that use the SMGR API could also be encrypted. This requires adding a marker access method and extending the event triggers, using the same approach as with heap tables.
+
+## Storage Manager (SMGR) API
```
dAdAbird, line 71: We also have some core mofdifications to make WAL encryption possible

documentation/docs/architecture/encryption-architecture.md
Andriciuc: It is indeed, thank you for pointing that out, I will sweep the current doc to make sure all lists do not have full stops.

documentation/docs/architecture/encryption-architecture.md
Andriciuc: True, confusing and providing no real information, deleted.

documentation/docs/advanced-topics/usage-guide.md
Andriciuc: Indeed, it is confusing, rewrote to make it clearer to the user that it is intended to run while PG server is stopped.

documentation/docs/advanced-topics/tech-reference.md
Andriciuc: Thank you, and no worries, it makes sense. I updated the icon to show :material-database-arrow-right:

documentation/docs/advanced-topics/usage-guide.md
Andriciuc: They do! Researching and updated this with proper arguments and an example arg: 

For keys and providers administration, it provides two functions:

```sql
pg_tde_GRANT_database_key_management_TO_role(role_name)
pg_tde_REVOKE_database_key_management_FROM_role(role_name)
```

These functions take a role name as a string argument, for example `'user1'`.

documentation/docs/advanced-topics/usage-guide.md
Andriciuc: Thank you for catching this, removed the  SET ACCESS METHOD tde_heap; arg is no-op because the table is already encrypted.

documentation/docs/architecture/encryption-architecture.md
Andriciuc: Hmm, just above SMGR I added this WAL mention and pointing the users to the config:

WAL encryption is implemented through a separate, server-wide mechanism and does not rely on table access methods or the SMGR API. For details, see [Configure WAL encryption](../wal-encryption.md).

Saying core modifications maybe is not the best approach, let me know if it's okay!

dAdAbird: "a slightly modified version of the SMGR API" is in fact the core changes that makes possible to encrypt everything that we encrypt now with the exception of WAL. To enable WAL encryption we have some API extension (core changes) as well. It's just seems a bit off to me that we mention one (SMGR) and skip on the other (WAL)

documentation/docs/architecture/encryption-architecture.md
```diff
@@ -0,0 +1,84 @@
...
+* Making the API generally extensible, where extensions can inject custom code into the storage manager
+* Adding tracking information for files. When a new file is created for an existing relation, references to the existing file are also passed to the SMGR functions
+
+With these modifications, `pg_tde` implements an additional layer on top of the normal Magnetic Disk SMGR API: if the related table is encrypted, `pg_tde` encrypts a file before writing it to the disk and, similarly, decrypts it after reading when needed.
+
+## WAL encryption
+
+WAL encryption is implemented through a separate, server-wide mechanism that extends PostgreSQL WAL-related APIs. Similar to the SMGR integration, this relies on additional extensions to PostgreSQL core APIs to support encryption of WAL records.
+
+For details, see [Configure WAL encryption](../wal-encryption.md).
```
dAdAbird: To me it reads as I can find out more about WAL-related core changes in `../wal-encryption.md`

PR #495, opened 2025-12-11
PG-2087 - Update the pg_tde installation steps and pg_tde availability table

Andriciuc:
This PR updates the commands to install pg_tde for apt and yum and informs the user that after PPG 17.7 they need to install the package manually.

### DIFF
--- a/documentation/docs/install.md
+++ b/documentation/docs/install.md
@@ -11,7 +11,7 @@ You can select from multiple easy-to-follow installation options to install `pg_
 | 18.x (and later)  | No | Install percona-pg-tde-(pg-version) package manually |
 
 !!! important
-    Starting with PPG 17.7, `pg_tde` is no longer bundled with the PostgreSQL server package. It is now delivered as a separate package named `percona-pg-tde-(pg-version)`.
+    Starting with PPG 17.7, `pg_tde` is no longer bundled with the Percona server for PostgreSQL package. It is now delivered as a separate package named `percona-pg-tde-(pg-version)`.
 
 !!! note
     Specific information on the supported platforms, products, and versions is described in the [Percona Software and Platform Lifecycle :octicons-link-external-16:](https://www.percona.com/services/policies/percona-software-support-lifecycle) page.


Andriciuc:
I have added two small link fixes in install.md and faq.md, unrelated to the packaging issue.

The rest of the files have been updated based on the recent packaging changes.

documentation/docs/install.md
```diff
@@ -4,10 +4,24 @@ You can select from multiple easy-to-follow installation options to install `pg_
...
+| **PostgreSQL version** | **Is pg_tde installed automatically?** | **Action Required** |
 | -------- | -------- | -------- |
-| 17.x (minor updates) | Yes | None |
-| 18.x (or later)  | No | Install package manually |
+| 17.x - 17.6 (minor updates) | Yes | None |
+| 17.7 (and later) | No | Install percona-pg-tde-(pg-version) package manually |
+| 18.x (and later)  | No | Install percona-pg-tde-(pg-version) package manually |
+
+!!! important
+    Starting with PPG 17.7, `pg_tde` is no longer bundled with the PostgreSQL server package. It is now delivered as a separate package named `percona-pg-tde-(pg-version)`. If you previously installed `pg_tde` from either the old bundled package `percona-postgresql-17-pg-tde`, or an earlier standalone package `percona-pg-tde-(pg-version)`, complete the following steps before installing the new version:
```
AndersAstrand: Should this be "is no longer bundled with the Percona server for PostgreSQL package"?

OR do we generally just refer to it as "the PostgreSQL server"?

documentation/docs/install.md
```diff
@@ -4,10 +4,24 @@ You can select from multiple easy-to-follow installation options to install `pg_
...
+
+    1. Drop the extension using `DROP EXTENSION … CASCADE`.
```
AndersAstrand: I thought we agreed to not encourage this?  Almost no-one is going to want to drop all their encrypted tables as a part of an upgrade procedure

AndersAstrand (changes requested):
I think the `DROP EXTENSION ... CASCADE` should be removed from the docs as per our previous discussion on the topic. It's almost never what anyone wants to do.

documentation/docs/install.md
Andriciuc: We generally refer to it by the longer format, Percona server for PostgreSQL. This also appears in release notes, FAQ's and other places. Nice catch, thank you for reminding me about the full name conv!

documentation/docs/install.md
Andriciuc: Ah I have removed this and kept only the initial informative note on the package name change.

PR #513, opened 2026-02-10
Somewhat fix readme

AndersAstrand:
The downloads section was pure fiction and the overview didn't include that psp18 is supported.


Andriciuc (approved):
LGTM!

PR #525, opened 2026-03-13
PG-1805 PG-1806 Do not overwrite existing SMGR keys on WAL replay

AndersAstrand:
In some cases postgresql will replay the key creation record from wal during crash recovery while there is already encrypted data on disk. This would make the data unreadable since the key it was encrypted with would be gone.

~~Two questions for reviewer:~~
- ~~Do we need to do the search for key and addition of the new one in a single pass instead of two?~~
- ~~Should we add tests for this to the pg_tde test suite, or is it being tested in the PSP testsuite enough?~~

Another potential solution is to include the encrypted key data in the WAL record, but I'm worried that would open a can of worms my gut tells me we don't want to open.

### DIFF
--- a/src/access/pg_tde_tdemap.c
+++ b/src/access/pg_tde_tdemap.c
@@ -88,13 +88,25 @@ static void pg_tde_write_one_map_entry(int fd, const TDEMapEntry *map_entry, off
 static int	pg_tde_file_header_write(const char *tde_filename, int fd, const TDESignedPrincipalKeyInfo *signed_key_info, off_t *bytes_written);
 static void pg_tde_initialize_map_entry(TDEMapEntry *map_entry, const TDEPrincipalKey *principal_key, const RelFileLocator *rlocator, const InternalKey *rel_key_data);
 static int	pg_tde_open_file_write(const char *tde_filename, const TDESignedPrincipalKeyInfo *signed_key_info, bool truncate, off_t *curr_pos);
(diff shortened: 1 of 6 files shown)


jeltz (dismissed):
This looks good to me. I am not the biggest fan of the re-use of keys but I do not see a cleaner way to do it unless we get really invasive and maybe not even then.

src/access/pg_tde_tdemap.c
```diff
@@ -105,7 +117,54 @@ pg_tde_save_smgr_key(RelFileLocator rel, const InternalKey *rel_key_data)
...
+			entry_already_exists = true;
+			write_offset = entry_offset;
+			break;
+		}
+
+		if (write_offset == 0 && entry.type == MAP_ENTRY_TYPE_EMPTY)
+			write_offset = entry_offset;
+	}
+
+	if (!entry_already_exists || replace_existing)
```
dutow, line 159: I'm a little worried if there's some corner case with this that could weaken our encryption for manipulation during redo, on the other hand I can't really think of any real scenarios and we are using GCM, which makes it even harder.

dutow (approved):

src/access/pg_tde_tdemap.c
AndersAstrand: Yeah I was also a bit worried about that, but couldn't see a better solution.

PR #540, opened 2026-03-18
PG-2252 Do not keep stale status for relations after smgr_close

AndersAstrand:
One backend might use its entry in the relcache for a relation that has been completely replaced by something else by another backend. This means that we no longer know the encryption status of this relation after smgr_close has been called.

This patch replaces the RELATION_KEY_NOT_AVAILABLE status with RELATION_ENCRYPTION_UNKNOWN and makes sure to lazily resolve the status on I/O.

### DIFF
--- a/t/reuse_relfilenode_in_cache.pl
+++ b/t/reuse_relfilenode_in_cache.pl
@@ -1,0 +1,154 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use File::Basename;
+use Test::More;
+use lib 't';
+use pgtde;
+
+PGTDE::setup_files_dir(basename($0));
+
+unlink('/tmp/reuse_relfilenode_in_cache.per');
+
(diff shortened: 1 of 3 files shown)


t/reuse_relfilenode_in_cache.pl
```diff
@@ -0,0 +1,154 @@
...
+$bg{run} = IPC::Run::start(
+	[
+		'psql', '--no-psqlrc', '--no-align',
+		'--file' => '-',
+		'--dbname' => $node->connstr('postgres')
+	],
+	'<' => \$bg{stdin},
+	'>' => \$bg{stdout},
+	'2>' => \$bg{stderr},
+	$psql_timeout);
```
jeltz, line 73: Maybe we should use `background_psql` instead here?

jeltz (approved):

PR #554, opened 2026-03-28
PG-2278 Re-use cipher contexts for SMGR encryption

jeltz:
Profiling with `perf` revealed that `EVP_CipherInit_ex()` showed up quite a lot when encrypting and decrypting tables, so to decrease how much it shows up we make sure to only allocate and initialize contexts for SMGR encryption once on loading `pg_tde` and then re-use those contexts but re-initializing them with different keys and IVs on use, which is a lot cheaper than doing a full initialization. This way of calling  `EVP_CipherInit_ex()` is poorly documented in OpenSSL.

WAL encryption already has logic for reusing contexts so this optimization is not as relevant there.

### DIFF
--- a/src/encryption/enc_aes.c
+++ b/src/encryption/enc_aes.c
@@ -32,26 +32,42 @@
  * 16 byte blocks.
  */
 
-static const EVP_CIPHER *cipher_cbc_128 = NULL;
 static const EVP_CIPHER *cipher_gcm_128 = NULL;
 static const EVP_CIPHER *cipher_ctr_ecb_128 = NULL;
 
-static const EVP_CIPHER *cipher_cbc_256 = NULL;
 static const EVP_CIPHER *cipher_gcm_256 = NULL;
 static const EVP_CIPHER *cipher_ctr_ecb_256 = NULL;
+
+static EVP_CIPHER_CTX *ctx_cbc_128 = NULL;
+static EVP_CIPHER_CTX *ctx_cbc_256 = NULL;
+
+static EVP_CIPHER_CTX *
+AesCbcInitCtx(const EVP_CIPHER *cipher, const char *name)
+{
+	EVP_CIPHER_CTX *ctx = EVP_CIPHER_CTX_new();
+
+	if (EVP_CipherInit_ex(ctx, cipher, NULL, NULL, NULL, 1) == 0)
+		ereport(ERROR,
+				errmsg("EVP_CipherInit_ex of %s failed. OpenSSL error: %s",
+					   name, ERR_error_string(ERR_get_error(), NULL)));
+
+	EVP_CIPHER_CTX_set_padding(ctx, 0);
+
+	return ctx;
+}
 
 void
 AesInit(void)
 {
 	OPENSSL_init_crypto(OPENSSL_INIT_ADD_ALL_CIPHERS | OPENSSL_INIT_LOAD_CRYPTO_STRINGS, NULL);
 
-	cipher_cbc_128 = EVP_aes_128_cbc();
 	cipher_gcm_128 = EVP_aes_128_gcm();
 	cipher_ctr_ecb_128 = EVP_aes_128_ecb();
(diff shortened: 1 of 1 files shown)


src/encryption/enc_aes.c
```diff
@@ -52,6 +55,18 @@ AesInit(void)
 	cipher_cbc_256 = EVP_aes_256_cbc();
 	cipher_gcm_256 = EVP_aes_256_gcm();
 	cipher_ctr_ecb_256 = EVP_aes_256_ecb();
+
+	ctx_cbc_128 = EVP_CIPHER_CTX_new();
+	if (EVP_CipherInit_ex(ctx_cbc_128, cipher_cbc_128, NULL, NULL, NULL, 1) == 0)
+		ereport(ERROR,
+				errmsg("EVP_CipherInit_ex failed. OpenSSL error: %s", ERR_error_string(ERR_get_error(), NULL)));
```
dutow: If we are copy-pasting this twice, shouldn't at least these use two distinct error messages?

jeltz: Yeah, we should. Or we should turn it into a function.

jeltz: I did both.

src/encryption/enc_aes.c
```diff
@@ -89,44 +101,39 @@ AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, int key_len, co
...
-	EVP_CIPHER_CTX *ctx = NULL;
-	const EVP_CIPHER *cipher;
+	EVP_CIPHER_CTX *ctx;
 
 	Assert(key_len == 16 || key_len == 32);
-	cipher = key_len == 32 ? cipher_cbc_256 : cipher_cbc_128;
-
-	Assert(cipher != NULL);
-	Assert(in_len % EVP_CIPHER_block_size(cipher) == 0);
+	ctx = key_len == 32 ? ctx_cbc_128 : ctx_cbc_128;
```
dutow: One should be 256

src/encryption/enc_aes.c
```diff
@@ -89,44 +101,39 @@ AesEcbEncrypt(EVP_CIPHER_CTX **ctxPtr, const unsigned char *key, int key_len, co
...
 		ereport(ERROR,
 				errmsg("EVP_CipherInit_ex failed. OpenSSL error: %s", ERR_error_string(ERR_get_error(), NULL)));
 
-	EVP_CIPHER_CTX_set_padding(ctx, 0);
-
 	if (EVP_CipherUpdate(ctx, out, &out_len, in, in_len) == 0)
 		ereport(ERROR,
 				errmsg("EVP_CipherUpdate failed. OpenSSL error: %s", ERR_error_string(ERR_get_error(), NULL)));
 
-	if (EVP_CipherFinal_ex(ctx, out + out_len, &out_len_final) == 0)
```
dutow, line 118: This and the padding disabling removal before together is definitely wrong.

By default openssl wants to write a padding block after each data, so we are misusing the API and silently skipping it, relying on implementation details.

If you keep `EVP_CIPHER_CTX_set_padding`, we should be able to skip `EVP_CipherFinal_ex` in practice, as it's a noop, but it is still part of the openssl api, so do we really want to?

src/encryption/enc_aes.c
jeltz: Wow, a bit scary that the code worked but I guess testing this is tricky.

src/encryption/enc_aes.c
jeltz: I am not sure, the issue here is really that if OpenSSL would write a padding block we would get a buffer overflow so if Final would ever do anything we would be totally hosed.

jeltz: Re-added the fianlize function since it is porbably cheap, but the set_padding function was never removed.

src/encryption/enc_aes.c
jeltz: Could sadly not figure out a reasonable way to test it.

dutow (approved):

PR #558, opened 2026-03-31
Update the changing database default tablespace

Andriciuc:
Update the database default tablespace behavior description to include alter/set command

Previously the explanation was not quite helpful, as in the user can technically still use the command, previously we said it was not supported which is not true.


documentation/docs/index/tde-limitations.md
```diff
@@ -49,16 +49,15 @@ Limitations of `pg_tde` {{release}}:
...
+    The `ALTER DATABASE ... SET TABLESPACE` command bypasses PostgreSQL's storage manager (SMGR), which `pg_tde` relies on to enforce encryption.
 
-    This operation bypasses PostgreSQL's storage manager (SMGR), which is not supported by `pg_tde`.
+    - If encrypted objects exist in the database's default tablespace, the operation is refused.
+    - If no encrypted objects are present in the default tablespace, the operation is allowed.
 
-    As a safeguard, `pg_tde` blocks the operation if encrypted objects are detected in the default tablespace.
-
-    Objects located outside the default tablespace are not affected by this command.
+    Only objects in the default tablespace are checked. Objects in other tablespaces are not affected by this validation.
```
jeltz: You should add the suggestion that you can move the encrypted tables one by one.

jeltz (changes requested):

documentation/docs/index/tde-limitations.md
Andriciuc: Updated, added that users can use ALTER TABLE SET TABLESPACE to move tables individually, should we also suggest the ALTER INDEX SET TABLESPACE command?

documentation/docs/index/tde-limitations.md
```diff
@@ -49,16 +49,17 @@ Limitations of `pg_tde` {{release}}:
...
 
-    This operation bypasses PostgreSQL's storage manager (SMGR), which is not supported by `pg_tde`.
+    - If encrypted objects exist in the database's default tablespace, the operation is refused.
+    - If no encrypted objects are present in the default tablespace, the operation is allowed.
 
-    As a safeguard, `pg_tde` blocks the operation if encrypted objects are detected in the default tablespace.
+    Only objects in the default tablespace are checked. Objects in other tablespaces are not evaluated by `pg_tde`.
 
-    Objects located outside the default tablespace are not affected by this command.
+  To move encrypted tables individually, use `ALTER TABLE ... SET TABLESPACE`.
```
jeltz: A bit weird way to phrase it. Maybe it should be more like: "A workaround is to ...".

jeltz (changes requested):

documentation/docs/index/tde-limitations.md
Andriciuc: Mmm, I see the idea but we do not want to give the impression this is a second hand solution, in this case it is a correct way to move encrypted tables safely. However, I did explain this a bit and added it under the warning so we have a proper continuation.

jeltz (approved):

PR #559, opened 2026-03-31
PG-2234 Fix pg_rewind with encrypted tables

dAdAbird:
The rewind might update blocks on target's file with that of the source.
In that case, encrypted files will end up with different blocks
encrypted with different internal keys, leading to data corruption.
To prevent this, we have to decryp blocks coming from the source and
re-encrypt them with the target's internal key.

Along with that, providers might have been changed, primary keys
rotated, etc., after the divergence of clusters. Generally, it's not a
problem because we would replace the keys and providers on the target
with those of the source. But it will render internal keys that we have
used for the partial blocks unreadable. To fix this, we have to
re-encrypt such internal keys with source's primary keys.

So the sequence is next:
1. Copy source's pg_tde dir into a tmp location. All following reads
and writes of source keys and providers happen in this tmp dir.
2. When we have partial updates, check for the source key. And
re-encrypt blocks if the key exists.
3. Save the target key into the source _keys file replacning the
existing one.
4. Replace target's pg_tde dir with the tmp one.

See commit messages for additional info.

There is no need to do anything with WAL keys. Although rewind
may keep a segment from the destination, it might only be a segment
that was created before common history (before the replica was created);
therefore, the key for such a segment would be copied on the replica
while pg_basebackup. The rest of the segments would be wiped out from
the destination and replaced by the source's ones (along with their keys).

All changes only for pg18 so far. I'll port them to pg17 when I get approvals


src/encryption/enc_tde.c
```diff
@@ -141,3 +141,67 @@ pg_tde_stream_crypt(const char *iv_prefix,
 		batch_no++;
 	}
 }
+
+/*
+ * The intialization vector of a block is its block number conmverted to a
```
artemgavrilov:
```suggestion
 * The initialization vector of a block is its block number converted to a
```

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,299 @@
...
+	unsigned int segNo;
+} current_file_data;
+
+static current_file_data current_tde_file =
+{
+	0
+};
+
+/* Dir for an operational copy of source's tde files (_keys, etc)  */
+static char tde_tmp_scource[MAXPGPATH] = "/tmp/pg_tde_rewindXXXXXX";
```
artemgavrilov:
```suggestion
static char tde_tmp_source[MAXPGPATH] = "/tmp/pg_tde_rewindXXXXXX";
```

fetools/pg18/pg_rewind/libpq_source.c
```diff
@@ -682,3 +718,52 @@ libpq_destroy(rewind_source *source)
...
+	PGresult   *res;
+
+	res = PQexec(conn, "SELECT pg_ls_dir('" PG_TDE_DATA_DIR "', true, false)");
+
+	if (PQresultStatus(res) != PGRES_TUPLES_OK)
+		pg_fatal("could not fetch file list: %s",
+				 PQresultErrorMessage(res));
+
+	/* no tde dir, nothing to do */
+	if (PQnfields(res) == 0)
```
artemgavrilov: Mmm... it seems that it should be `PQntuples` (rows), not `PQnfields` (columns) here

fetools/pg18/pg_rewind/filemap.c
```diff
@@ -572,9 +572,33 @@ isRelDataFile(const char *path)
...
 	unsigned int segNo;
-	int			nmatch;
 	bool		matched;
 
+	matched = path_rlocator(path, &rlocator, &segNo);
+	if (matched)
+	{
+		char	   *check_path = datasegpath(rlocator, MAIN_FORKNUM, segNo);
+
+		if (strcmp(check_path, path) != 0)
```
jeltz, line 593: Seems like you accidentally removed the comment explaining this part of the code.

fetools/pg18/pg_rewind/filemap.c
```diff
@@ -572,9 +572,33 @@ isRelDataFile(const char *path)
...
+		pfree(check_path);
+	}
+
+	return matched;
+}
+
+/*
+ * Sets rlocator and segNo based on given path. Returns false if didn't find
+ * a match.
+ *
```
jeltz: Trailing whitespace.

fetools/pg18/pg_rewind/filemap.c
```diff
@@ -712,6 +719,15 @@ decide_file_action(file_entry_t *entry)
 	if (strstr(path, ".DS_Store") != NULL)
 		return FILE_ACTION_NONE;
 
+	/*
```
jeltz: Trailing whitespace.

fetools/pg18/pg_rewind/pg_rewind.c
```diff
@@ -600,6 +600,7 @@ perform_rewind(filemap_t *filemap, rewind_source *source,
 			while (datapagemap_next(iter, &blkno))
 			{
 				offset = blkno * BLCKSZ;
+
```
jeltz: Accidental extra change.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,216 @@
...
+	InternalKey *target_key;
+	char		path[MAXPGPATH];
+	RelFileLocator rlocator;
+	unsigned int segNo;
+} current_file_data;
+
+static current_file_data current_tde_file =
+{
+	0
+};
```
jeltz: I think we usually write these kind of zero initializers as just `{0}` on the same line as the delcaration.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,216 @@
...
+
+/* Dir for an operational copy of source's tde files (_keys, etc)  */
+static char tde_tmp_scource[MAXPGPATH] = "/tmp/pg_tde_rewindXXXXXX";
+
+void
+flush_current_key(void)
```
jeltz: This should be named something related to tde.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,216 @@
...
+
+		memset(current_tde_file.path, 0, MAXPGPATH);
+		strlcpy(current_tde_file.path, relpath, MAXPGPATH);
+		current_tde_file.rlocator = rlocator;
+		current_tde_file.segNo = segNo;
+	}
+}
+
+void
+encrypt_block(unsigned char *buf, off_t file_offset)
```
jeltz: This should be named something with tde.

jeltz (changes requested):
Just some quick review comments.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,299 @@
...
+
+		memset(current_tde_file.path, 0, MAXPGPATH);
+		strlcpy(current_tde_file.path, relpath, MAXPGPATH);
+		current_tde_file.rlocator = rlocator;
+		current_tde_file.segNo = segNo;
+	}
+}
+
+void
+encrypt_block(unsigned char *buf, off_t file_offset, ForkNumber fork)
```
artemgavrilov: It seems that func should be named `reencrypt`

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,299 @@
...
+	for (;;)
+	{
+		ssize_t		read_len;
+
+		read_len = read(srcfd, buf.data, sizeof(buf));
+
+		if (read_len < 0)
+			pg_fatal("could not read file \"%s\": %m", srcpath);
+		else if (read_len == 0)
+			break;				/* EOF reached */
```
artemgavrilov: I think we should also check that we read entire block here, because later in crypt/decrypt we expect that we have exactly BLCKSZ bytes.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,299 @@
...
+		close(srcfd);
+		return;
+	}
+
+	written_len = 0;
+	for (;;)
+	{
+		ssize_t		read_len;
+
+		read_len = read(srcfd, buf.data, sizeof(buf));
```
artemgavrilov: `sizeof(buf)` is bigger than `buf.data`. It should be just BLCKSZ I think.

artemgavrilov (commented):
Partial review comments

fetools/pg18/pg_rewind/local_source.c
```diff
@@ -170,12 +187,37 @@ local_queue_fetch_range(rewind_source *source, const char *path, off_t off,
...
+static bool
+directory_exists(const char *dir)
+{
+	struct stat st;
+
+	if (stat(dir, &st) != 0)
+		return false;
+	if (S_ISDIR(st.st_mode))
+		return true;
+	return false;
```
jeltz: I think this could be simplified a bit as `return stat(dir, &st)  == 0 && S_ISDIR(st.st_mode);`.

fetools/pg18/pg_rewind/local_source.c
```diff
@@ -170,12 +187,37 @@ local_queue_fetch_range(rewind_source *source, const char *path, off_t off,
...
+	create_tde_tmp_dir();
+	copy_tmp_tde_files(tde_source_dir);
+}
+
 static void
 local_finish_fetch(rewind_source *source)
 {
-	/*
-	 * Nothing to do, local_queue_fetch_range() copies the ranges immediately.
-	 */
```
jeltz, line 178: Maybe write a comment replacing this.

src/access/pg_tde_tdemap.c
```diff
@@ -164,10 +162,15 @@ pg_tde_save_smgr_key(RelFileLocator rel,
 		pg_tde_write_one_map_entry(fd, &write_entry, &write_offset, file_path);
 	}
 
+#ifdef FRONTEND
+	pfree(principal_key);
+#endif
```
jeltz, line 167: Is this really realted to this commit?

src/encryption/enc_tde.c
```diff
@@ -141,3 +141,67 @@ pg_tde_stream_crypt(const char *iv_prefix,
...
+
+void
+tde_encrypt_smgr_block(InternalKey *relKey, ForkNumber forknum, BlockNumber blocknum, const unsigned char *in, unsigned char *out)
+{
+	unsigned char iv[16];
+
+	CalcBlockIv(forknum, blocknum, relKey->base_iv, iv);
+
+	AesEncrypt(relKey->key, relKey->key_len, iv, in, BLCKSZ, out);
+}
```
jeltz, line 207: Maybe this change should be in a separate commit since this refactoring seems to make sense anyway. Turning it more int a library.

jeltz (changes requested):

jeltz:
The second commit message needs to be improved and the first contains some typos, e.g. "blok" instead of "block".

fetools/pg18/pg_rewind/libpq_source.c
```diff
@@ -354,6 +364,17 @@ libpq_queue_fetch_file(rewind_source *source, const char *path, size_t len)
 static void
 libpq_queue_fetch_range(rewind_source *source, const char *path, off_t off,
```
jeltz, line 365: Why have this wrapper instead of just changing all callsites to pass either true or false?

fetools/pg18/pg_rewind/local_source.c
```diff
@@ -210,7 +210,7 @@ local_fetch_tde_keys(rewind_source *source)
 	if (!directory_exists(tde_source_dir))
 		return;
 
-	create_tde_tmp_dir();
+	init_tde();
```
jeltz, line 209: Maybe this change should be moved to the first commit?

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -195,6 +200,14 @@ copy_dir(const char *src, const char *dst)
 		pg_fatal("could not close directory \"%s\": %m", src);
 }
 
+void
+init_tde(void)
+{
+	source_has_tde = true;
+	create_tde_tmp_dir();
+	atexit(destroy_tde_tmp_dir);
+}
```
jeltz, line 357: Same with this. Maybe `atexti()` should be in the frist commit?

t/pg_rewind_enc_copy_blocks.pl
```diff
@@ -0,0 +1,82 @@
+
```
jeltz: Remove this extra newline.

t/pg_rewind_enc_copy_blocks.pl
```diff
@@ -0,0 +1,82 @@
...
+# Copyright (c) 2021-2024, PostgreSQL Global Development Group
```
jeltz: We generally do not add copyright headers to our files.

fetools/pg18/pg_rewind/filemap.c
```diff
@@ -719,13 +721,11 @@ decide_file_action(file_entry_t *entry)
...
 
-	/* 
+	/*
 	 * Skip pg_tde key data but WAL-related stuff as WAL being replaced by
 	 * source's. We will handle the rest while re-encrypting data.
 	 */
-	if (strstr(path, "pg_tde/") != NULL &&
-		strstr(path, "pg_tde/wal_keys") == NULL &&
-		strstr(path, "pg_tde/1664_providers") == NULL)
+	if (strstr(path, "pg_tde/") != NULL)
```
jeltz, line 737: Why this change? Is it in the righyt commit? If so it is a bit confusing the comment did not change.

t/pg_rewind_enc_unchanged_rel.pl
```diff
@@ -0,0 +1,65 @@
+
+# Copyright (c) 2021-2024, PostgreSQL Global Development Group
```
jeltz: Same comment about empty line and copyright notice.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -35,12 +35,79 @@ static current_file_data current_tde_file =
 static char tde_tmp_scource[MAXPGPATH] = "/tmp/pg_tde_rewindXXXXXX";
 static bool source_has_tde = false;
 
+static void
+recrypt_fork(ForkNumber fork)
```
jeltz: I think it is "reencrypt".

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -35,12 +35,79 @@ static current_file_data current_tde_file =
...
+		return;
+
+	srcfd = open(srcpath, O_RDONLY | PG_BINARY, 0);
+	if (srcfd < 0)
+	{
+		/*
+		 * Server can recover from wrecked VM/FSM, hence only warnings here
+		 * and in the rest of the function
+		 */
+		pg_log_warning("could not open file for reading \"%s\": %m", srcpath);
```
jeltz: I think these error messages maybe should be improved with some context.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -35,12 +35,79 @@ static current_file_data current_tde_file =
...
+
+	written_len = 0;
+	for (;;)
+	{
+		ssize_t		read_len;
+
+		read_len = read(srcfd, buf.data, sizeof(buf));
+
+		if (read_len < 0)
+			pg_fatal("could not read file \"%s\": %m", srcpath);
```
jeltz: Why is this fatal and not jsut a warning? Doesn't PostgreSQL recreate them if they are corrupt?

t/pg_rewind_enc_fsm.pl
```diff
@@ -0,0 +1,68 @@
+
+# Copyright (c) 2021-2024, PostgreSQL Global Development Group
```
jeltz: Same issue with empty line and copyright.

fetools/pg18/pg_rewind/filemap.c
dAdAbird: The comment is in inside path_rlocator where the path parsing is

src/access/pg_tde_tdemap.c
dAdAbird: Yes, because we expose it this part to the FE tools. And sanitizers now complain on a memory leak w/o pfree

src/encryption/enc_tde.c
dAdAbird: Yeah, makes sense

fetools/pg18/pg_rewind/filemap.c
jeltz: No?

```
	/*
	 * The sscanf tests above can match files that have extra characters at
	 * the end. To eliminate such cases, cross-check that GetRelationPath
	 * creates the exact same filename, when passed the RelFileLocator
	 * information we extracted from the filename.
	 */
```

dAdAbird: Ah, I thought you are talking about another comment. Yes, indeed, this one got disappeared somehow. I'll fix, thanks!

fetools/pg18/pg_rewind/libpq_source.c
dAdAbird: This is what `rewind_source` [interface](https://github.com/percona/pg_tde/blob/1b1180b0a1cd8662a7bffa4971c89ffe9e1b82c1/fetools/pg18/pg_rewind/rewind_source.h#L47-L60) demands. For the  "local source", `queue_fetch_range` and `queue_fetch_range` are actually two different functions. I'm not eager to change this interface tbh, as it might bring additional pain with upstream changes.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -0,0 +1,318 @@
...
+				pg_log_warning("could not read block from fork file \"%s\": %m", srcpath);
+				pg_log_warning_hint("%s", warning_hint);
+			}
+
+			break;				/* EOF reached if read_len == 0 */
+		}
+
+		if (read_len != BLCKSZ)
+		{
+			pg_log_warning("unexpected read from fork file \"%s\": %m", srcpath);
```
artemgavrilov: `%m` here will may contain garbage as `errno` set only when `read` returns `-1`
https://man7.org/linux/man-pages/man2/read.2.html

artemgavrilov (approved):
Seems fine

fetools/pg18/pg_rewind/tde_ops.c
dAdAbird: eah, a copy-paste issue) thanks

Andriciuc (approved):
LGTM! Thank you for updating the variables!

jeltz:
Typo in second commit message:

>Along with that, providers might have been changed, primary keys rotated, etc., after the divergence of clusters.

It should say "principal keys".

dutow:
I did a claude-assisted test run on the changes, focusing only on trying to break rewind/recovery, not code/quality review. It identified the following issues (all verified with custom test scripts):

## 1. multi segment corruption

For any TDE relation > 1 GB where the relation was created after standby init, pg_rewind corrupts every segment after the first. Triggered by my t/test_multisegment.pl (600k rows, 1.14 GB).

Root cause in fetools/pg18/pg_rewind/tde_ops.c:
1. ensure_tde_keys("base/X/Y") for segment 0 → loads source_key = IK_S, target_key = IK_P from the respective key files.
2. Partial blocks of segment 0 → tde_reencrypt_block correctly decrypts with IK_S, encrypts with IK_P.
3. Transition to segment 1: ensure_tde_keys("base/X/Y.1") → flush_current_tde_rel_key() for segment 0:
    - pg_tde_save_smgr_key(rlocator, target_key=IK_P, replace_existing=true) against tde_tmp_source. The entry for this rlocator in tmp_source is now IK_P, not IK_S.
4. For segment 1, source_key = pg_tde_get_smgr_key(rlocator) reads from tmp_source → returns IK_P.
5. Partial blocks of segment 1 arrive from source encrypted with IK_S. Code decrypts with IK_P → garbage. Encrypts with IK_P → garbage on disk.

## 2. same corruption can happen with fsm/vm

and the fix will be likely the same, just mentioning it for completeness

## 3. --dry-run modifies the target

  diff -r before/ pgdata/ → Binary files .../base/5/16434_fsm differ

  tde_ops.c only checks dry_run inside fetch_tde_dir(). None of ensure_tde_wal_seg(), reencrypt_fork(), flush_current_tde_rel_key() (calls pg_tde_save_smgr_key), or tde_reencrypt_block() check it. With
  --dry-run:
  - ensure_tde_wal_seg rewrites kept WAL segments via pg_pwrite
  - reencrypt_fork rewrites FSM/VM forks
  - Tmp source dir is modified (less critical, gets cleaned up via atexit)

Even though the rewritten content is functionally equivalent (decrypt+re-encrypt with same key on same IV ⇒ identical ciphertext), zero-plaintext FSM pages become encrypted-zero ciphertext, so the file genuinely differs on disk. Violates --dry-run contract.

## 4. restored WAL segments left in target's pg_wal after pg_rewind with -c are encrypted with target's now-discarded keys

This is most likely the bug Mohit submitted as PG-2397.

After pg_rewind with -c against a pg_tde cluster using archive_command/restore_command, target startup fails with:

LOG:  invalid magic number 6A54 in WAL segment 000000010000000000000007
LOG:  redo done at 0/6068D88
FATAL: WAL ends before consistent recovery point

Root cause

1. pg_rewind runs extractPageMap after traverse_datadir. With -c, extractPageMap calls restore_command to pull target's old TLI-1 WAL segments back into pg_wal — but the filemap is already built, so
those segments get no entry (no REMOVE, no ENSURE_WAL_SEG).
2. The restore_command is pg_tde_restore_encrypt, which locates pg_tde via <dest>/../pg_tde — that's target's pg_tde --so it encrypts the restored WAL with target's WAL keys.
3. fetch_tde_dir() at the end of pg_rewind replaces target's pg_tde with source's. Target's WAL keys are now gone from the keyring.
4. Recovery scans WAL after the last common checkpoint, hits the segment 6→7 boundary, tries to read the page header of segment 7, decrypts it with the only key available (source's) → garbage → magic
6A54 instead of D11A → FATAL.

Trigger conditions (all four must hold)

- --restore-target-wal / -c on pg_rewind
- pg_tde archive tools as archive_command / restore_command
- Target's pg_wal has had post-divergence TLI-1 segments recycled out, so extractPageMap actually invokes restore_command
- Recovery touches one of those residual segments (always happens via the next-page-header lookahead at a segment boundary)


Note: it also produced several reports which were either caused by concurrent pg_tde operations on the live source, or trying to run rewind again after an aborted pg_rewind and crashing - I ignored those. My understanding that if pg_rewind fails at any time, we should expect the data directory to be completely broken, so we don't have to deal with that situation - but if I'm wrong about that, we might have a few issues there.

dAdAbird:
> My understanding that if pg_rewind fails at any time, we should expect the data directory to be completely broken

I agree

dAdAbird:
It is became too monstrous to process. So we decided to create a feature branch based on it and proceed with PR into that.
The last comment made into Jira ticket: https://perconadev.atlassian.net/browse/PG-2407

PR #567, opened 2026-04-09
Run OpenBao automatically in TAP tests

jeltz:
To make running the tests easier we let the TAP tests handle the starting, stopping and configuring of OpenBao. The plan is to do the same for OpenKMIP but let's do things step by step.

In the future we may want to add the ability to run the test suite against e.g. Hashicorp Vault Enterprise. For that we could for example add an environment variable which tells the tests to connect an existing server rather than spawning a new one. But let's cross that bridge when we get to it.

**Question for the reviewers:** My plan is to turn this into a Perl module, add support for random ports and turn the Vault tests into a TAP test but I decided to start with just this step. Is this initial step too small?

### DIFF
--- a/t/regress.pl
+++ b/t/regress.pl
@@ -66,7 +66,9 @@ sub bao_setup
 	$bao_pid = fork;
 	unless ($bao_pid)
 	{
-		exec('bao', 'server', '-dev', '-dev-tls',
+		exec(
+			'bao', 'server', '-dev', '-dev-tls',
+			"-dev-listen-address=127.0.0.1:$bao_port",
 			"-dev-cluster-json=$bao_temp/info");
 	}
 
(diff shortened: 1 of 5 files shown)


t/regress.pl
```diff
@@ -51,3 +57,79 @@ BEGIN
   or exit $?;
 
 $node->stop;
+
+sub bao_setup
+{
+	my $bao_temp = PostgreSQL::Test::Utils::tempdir('bao');
+	my $bao_port = 8200;
+
+	$bao_pid = fork;
```
dutow: Wouldn't it be better to use IPC::Run::start?

dutow (commented):
One comment/question, otherwise looks good

dutow (approved):

PR #572, opened 2026-04-23
PG-2308 Add server principal key cache for frontend code

artemgavrilov:
PG-2308

pg_tde_basebackup reads server principal key multiple times during its run. At the same time it copies key providers configs and overwrites that ones that were pre-copied by user. That leads to race condition when basebackup tries to read file that in the middle of overwriting from another child process.

This PR adds server key cache for frontend code so server key will be read only once, before `fork()`.

### DIFF
--- a/src/catalog/tde_principal_key.c
+++ b/src/catalog/tde_principal_key.c
@@ -934,6 +934,13 @@ pg_tde_get_key_info(PG_FUNCTION_ARGS, Oid dbOid)
 
 #endif							/* FRONTEND */
 
+#ifdef FRONTEND
+/*
+ * Process-local cache for the server (GLOBAL_DATA_TDE_OID) principal key.
+ */
+static TDEPrincipalKey *fe_server_principal_key_cache = NULL;
+#endif							/* FRONTEND */
+
 /*
  * Get principal key form the keyring.
  */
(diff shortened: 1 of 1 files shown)


src/catalog/tde_principal_key.c
```diff
@@ -1050,6 +1068,16 @@ GetPrincipalKeyNoDefault(Oid dbOid, LWLockMode lockMode)
...
 		principalKey = get_principal_key_from_cache(dbOid);
 	}
+#else
+	/* Cache on first successful read for GLOBAL_DATA_TDE_OID */
+	if (principalKey != NULL &&
+		dbOid == GLOBAL_DATA_TDE_OID &&
+		fe_server_principal_key_cache == NULL)
+	{
+		fe_server_principal_key_cache = palloc_object(TDEPrincipalKey);
+		if (fe_server_principal_key_cache)
```
jeltz: This cannot happen since `palloc_object()` calls `exit()` in frontend code.

jeltz (changes requested):

src/catalog/tde_principal_key.c
artemgavrilov: Ah, right. Removed this check.

jeltz (approved):

PR #575, opened 2026-05-04
Switch KMIP testing to cosmian

dutow:
PyKMIP seems to be abandonware and requires an older python version. This commt switches instead to use a Cosmian based kmip test, which also moves the server setup inside the TAP test to simplify local deployment.

Test assumes that the required binaries are installed and are user executable.

(Should we use something other than perl for testing?)


ci_scripts/ubuntu-deps.sh
```diff
@@ -51,7 +51,13 @@ sudo apt-get update
...
 
 sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
-pip3 install pykmip
+
+# Cosmian KMS
+wget https://package.cosmian.com/kms/5.21.0/deb/$(dpkg --print-architecture)/non-fips/static/cosmian-kms-server-non-fips-static-openssl_5.21.0_$(dpkg --print-architecture).deb
+sudo dpkg -i cosmian-kms-server-non-fips-static-openssl_5.21.0_$(dpkg --print-architecture).deb
+# .deb ships binary + bundled legacy.so as 0500 root:root; CI runner is non-root.
+sudo chmod 0755 /usr/sbin/cosmian_kms
+sudo chmod 0755 /usr/local/cosmian/lib/ossl-modules/legacy.so
```
jeltz, line 60: Maybe something you should report in their bug tracker?

ci_scripts/ubuntu-deps.sh
```diff
@@ -51,7 +51,13 @@ sudo apt-get update
 sudo apt-get install -y ${DEPS[@]}
 
 sudo bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)"
-pip3 install pykmip
+
+# Cosmian KMS
+wget https://package.cosmian.com/kms/5.21.0/deb/$(dpkg --print-architecture)/non-fips/static/cosmian-kms-server-non-fips-static-openssl_5.21.0_$(dpkg --print-architecture).deb
+sudo dpkg -i cosmian-kms-server-non-fips-static-openssl_5.21.0_$(dpkg --print-architecture).deb
```
jeltz, line 57: Should we add a variable for the architecture?

jeltz (commented):
Just two small comments.

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+
+sub find_binary
+{
+	if (my $override = $ENV{COSMIAN_KMS_BIN})
+	{
+		return $override if -x $override;
+		diag("COSMIAN_KMS_BIN=$override is not executable; ignoring");
+	}
+
+	for my $dir (File::Spec->path)
```
jeltz, line 21: This seems overkill. Why not just check if it is in the current path?

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+	{
+		my $candidate = File::Spec->catfile($dir, 'cosmian_kms');
+		return $candidate if -x $candidate;
+	}
+	return undef;
+}
+
+sub _run_openssl
```
jeltz: You should probably use `system_or_bail` instead of inventing your own helper.

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+	# Cosmian 5.21.0 silently ignores CLI flags when any TOML config
+	# is present (the Ubuntu .deb installs /etc/cosmian/kms.toml and
+	# cosmian's `-c` flag's help text says "All other command line
+	# arguments and environment variables are ignored once the
+	# configuration file is loaded"). Always launch via -c <our toml>.
+	my @cmd = ($bin, '-c', "$dir/kms.toml");
+
+	# Cosmian needs the openssl legacy provider to read the P12 bundle.
+	# Search known module dirs across distros; on Ubuntu CI the post-install
+	# chmod in ci_scripts/ubuntu-deps.sh exposes cosmian's bundled copy.
```
jeltz: I don't really get why the code below is necessary and I do not really feel the comment explains it.

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+
+sub _spawn
+{
+	my ($bin, $dir, $stderr_ref) = @_;
+
+	# Cosmian 5.21.0 silently ignores CLI flags when any TOML config
+	# is present (the Ubuntu .deb installs /etc/cosmian/kms.toml and
+	# cosmian's `-c` flag's help text says "All other command line
+	# arguments and environment variables are ignored once the
+	# configuration file is loaded"). Always launch via -c <our toml>.
```
jeltz: This comment adds no value as far as I can see.

ci_scripts/ubuntu-deps.sh
jeltz, line 60: Why use legacy rather than fips?

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+		'x509', '-req',
+		'-in', "$dir/client.csr",
+		'-CA', "$dir/ca.pem",
+		'-CAkey', "$dir/ca.key",
+		'-CAcreateserial', '-days',
+		'1', '-out',
+		"$dir/client.pem");
+}
+
+sub _pick_free_port
```
jeltz: Why not use PostgreSQL's helper for this?

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+		remove_tree("$dir/db", { safe => 1 });
+	}
+	BAIL_OUT("cosmian_kms failed to become ready after 3 attempts.\n$diag");
+}
+
+sub stop
+{
+	my ($h) = @_;
+	return unless $h;
+	eval { $h->kill_kill };
```
jeltz: Why eval? Why do we need to catch errors here?

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return start(\@cmd, '>', '/dev/null', '2>', $stderr_ref);
+}
+
+sub start_with_free_port
+{
+	my ($bin, $dir) = @_;
+	my $diag = '';
+	for my $attempt (1 .. 3)
```
jeltz: Not sure this loop is necessary. PostgreSQL seems to just live with the potential race condition and that test suite works fine in practice.

t/CosmianKms.pm
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		if (_wait_ready($http_port))
+		{
+			return ($h, $kmip_port);
+		}
+
+		$diag .= "[attempt $attempt kmip=$kmip_port http=$http_port] "
+		  . "readiness timed out\nstderr:\n$stderr\n";
+		eval { $h->kill_kill };
```
jeltz: Why this paranoia about kill spefically failing?

t/kmip.pl
```diff
@@ -0,0 +1,137 @@
...
+# Cosmian + cert setup.
+# ---------------------------------------------------------------------------
+my $tmpdir = PostgreSQL::Test::Utils::tempdir();
+CosmianKms::gen_certs($tmpdir);
+my ($cosmian_h, $kmip_port) =
+  CosmianKms::start_with_free_port($cosmian_bin, $tmpdir);
+
+END
+{
+	CosmianKms::stop($cosmian_h) if $cosmian_h;
```
jeltz: I can't say I like that both `stop` itself and this block checks for if the pid is set- Just check it in one place.

t/kmip.pl
```diff
@@ -0,0 +1,137 @@
...
+		"SELECT pg_tde_add_database_key_provider_kmip("
+	  . "'kmip-prov', '127.0.0.1', $kmip_port, "
+	  . "'$tmpdir/client.pem', '$tmpdir/client.key', '$tmpdir/ca.pem');"
+	  . "SELECT pg_tde_create_key_using_database_key_provider("
+	  . "'kmip-key', 'kmip-prov');"
+	  . "SELECT pg_tde_set_key_using_database_key_provider("
+	  . "'kmip-key', 'kmip-prov');"
+	  . "CREATE TABLE test_enc("
+	  . "id SERIAL, k INTEGER NOT NULL, PRIMARY KEY(id)) USING tde_heap;"
+	  . "INSERT INTO test_enc(k) VALUES (1), (2), (3);");
```
jeltz: This code is really hard to read. Why don't you use a heredoc or multiline string?

t/kmip.pl
```diff
@@ -0,0 +1,137 @@
...
+is($node->safe_psql('postgres', 'SELECT k FROM test_enc ORDER BY id;'),
+	"1\n2\n3", 'decrypts after restart (KMIP re-fetch)');
+
+# Key rotation + post-rotation insert.
+$node->safe_psql('postgres',
+		"SELECT pg_tde_create_key_using_database_key_provider("
+	  . "'kmip-key2', 'kmip-prov');"
+	  . "SELECT pg_tde_set_key_using_database_key_provider("
+	  . "'kmip-key2', 'kmip-prov');"
+	  . "INSERT INTO test_enc(k) VALUES (4), (5);");
```
jeltz: This code is also very hard to read.

t/kmip.pl
```diff
@@ -0,0 +1,137 @@
...
+# Restart after rotation.
+$node->restart;
+is($node->safe_psql('postgres', 'SELECT k FROM test_enc ORDER BY id;'),
+	"1\n2\n3\n4\n5", 'rows readable after restart following rotation');
+
+# ---------------------------------------------------------------------------
+# Negative path: pg_tde fails fast (<= 5s) when KMIP endpoint accepts TCP
+# but immediately closes (no TLS, no KMIP).
+# ---------------------------------------------------------------------------
+sub _bind_and_fork_reject_listener
```
jeltz, line 113: Adding a new test like this should be done in a separate commit.

jeltz (changes requested):
I cannot say I was happy with doing this review. Feels like it would have needed quite a bit more work before it would actually have been ready for review.

t/CosmianKms.pm
dutow: Yeah, maybe I could make this comment better, basically cosmian-kms needs OPENSSL_MODULES set

t/CosmianKms.pm
dutow: Hm, maybe. I left it there because this was really annoying

t/CosmianKms.pm
dutow: Because perl doesn't have a built in for that, a short if seems a good alternative to me vs adding a new package dependency

t/CosmianKms.pm
dutow: I found it but because of its placement I thought it's part of the cluster setup infrastructure, not generic... will fix.

t/CosmianKms.pm
dutow: No, now that I looked at how its implemented this is also part of the postgres port selection function.

ci_scripts/ubuntu-deps.sh
dutow: Would that be more reliable than calling dpkg?

ci_scripts/ubuntu-deps.sh
dutow: We are using non fips mode everywhere, so this seemed to be more consistent

dutow: Legacy is misleading here, it just means it also includes non fips compliant algorithms

dutow: this is documented, not a bug, seems to be part of "security". But completely wrong for testing. Still downloadng the packages and chmoding seemed better than building it from source.

t/kmip.pl
dutow: done

ci_scripts/ubuntu-deps.sh
jeltz: No, my idea was jsut to reduce duplicated calls to dpkg.

ci_scripts/ubuntu-deps.sh
jeltz: Ok, yeah, that is one way to add "security".

jeltz: I do not think that kind of consistency matters either way. This is a separate piece of software.

t/CosmianKms.pm
jeltz: I don't really see how this comment adds anything. It is annoying but not really anything we need to document since they already do it.

jeltz (approved):
Nice to not have to use the broken Python server anymore!

jeltz (commented):
As a bonus it would be nice if you could also remove the action in Github which downgrades Python but if you wont do it I will in a separate PR.

PR #580, opened 2026-05-13
Add Percona Live banner for pg_tde

Andriciuc:
This PR adds the PL banner.


jeltz (approved):
We can just merge this but I think this raises an interesting general problem. Maybe we should separate the CSS and HTML templates from the actual markdown files? And the styling should be done in one central docs repo?

PR #581, opened 2026-05-19
PG-2379 Allow different principal keys for relation key file migration

AndersAstrand:
The code assumed that all databases would use the same principal key which is generally not true. This fix loads the principal key for each database as needed.


rjd15372 (approved):
LGTM

PR #582, opened 2026-05-19
PG-2381 Fix migration with empty files and entries

AndersAstrand:
The reproduction script in this issue surfaced two issues. One with an empty key file that was created by VACUUM FULL, and one with empty entries in the key file created by DROP TABLE. None of these issues had anything to do with pg_tde_upgrade but could be triggered even by only upgrading pg_tde without changing postgresql version at all.


rjd15372 (approved):
Please note that my knowledge about this is very limited. But the changes look fine.

artemgavrilov (approved):

PR #583, opened 2026-05-20
Update release note with warning mention for pg_tde 2.2.0

Andriciuc:
This PR adds to the release notes a warning regarding pg_tde 2.2.0 working only with 17.10 and 18.4

### DIFF
--- a/documentation/docs/release-notes/release-notes-v2.2.0.md
+++ b/documentation/docs/release-notes/release-notes-v2.2.0.md
@@ -8,6 +8,9 @@ The `pg_tde` extension, provided by Percona, adds [Transparent Data Encryption (
 
 `pg_tde` now supports 256-bit AES encryption and introduces [`pg_tde_upgrade`](../command-line-tools/pg-tde-upgrade.md), a utility that simplifies the upgrades of encrypted clusters. For more details, see the [Changelog](#changelog).
 
+!!! warning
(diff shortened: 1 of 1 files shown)


documentation/docs/release-notes/release-notes-v2.2.0.md
```diff
@@ -8,6 +8,9 @@ The `pg_tde` extension, provided by Percona, adds [Transparent Data Encryption (
 
 `pg_tde` now supports 256-bit AES encryption and introduces [`pg_tde_upgrade`](../command-line-tools/pg-tde-upgrade.md), a utility that simplifies the upgrades of encrypted clusters. For more details, see the [Changelog](#changelog).
 
+!!! warning
+    `pg_tde` 2.2.0 requires Percona Distribution for PostgreSQL 17.10 or 18.4. It is not compatible with earlier versions of the distribution.
```
AndersAstrand: Technically it doesn't require the distribution at all, so maybe we should say that it doesn't work with distributions older than this instead? OR maybe it's true enough and noone uses pg_tde without the distribution anyways.

Andriciuc: No, you are right, we need to mention it as older distributions, thank you for the correction! Updated.

AndersAstrand (approved):

PR #585, opened 2026-05-20
Add 256 bit ciphers to the FAQ entry

AndersAstrand:
The new FAQ entry is just copy pasted from the architecture chapter.


jeltz (approved):

Andriciuc (approved):
LGTM! Thank you for the update Andreas!

PR #588, opened 2026-05-22
Proof of concept for writing more idiomatic TAP tests

jeltz:
This tries to not clean anything up outside of making the test more idiomatic TAP to test how this works out.

Things I want to check:

- Do we need some helpers for assertions when running SQL queries?
- Does anyone have any suggestion for how to improve readability?
- Is this better than what we have?


dAdAbird:
It is better to me.
`PostgreSQL::Test::Utils::tempdir;`is definitely what we should switch to. As well as checking the result instead of comparing files.
Regarding helpers, I'm kinda on the fence. On one hand they can be quite handy. For example `check_query()` in `RewindTest` compares output w/o formatting. Plus hides all the requests and "is()'. That makes the test much cleaner and easier to comprehend. On the other hand helpers might be not to flexible (or to complicated otherwise). For example you can't specify db_name for check_query(). So I don't know. I'm inside a bit more toward helper though, just for a sake of reducing visual noise in the test

PR #590, opened 2026-05-27
Make TAP tests for basic SMGR usage more idiomatic

jeltz:
The pattern with trying to emulate pg_regress using TAP where we save the output of psql to a text file has proven hard to maintain. And especially the tests which read the relation files become extra tricky to understand.

This initentionally leaves, and exposes, some quite ugly things but those can be fixed to future.


t/basic.pl
```diff
@@ -2,25 +2,23 @@
 
 use strict;
 use warnings;
-use File::Basename;
+use PostgreSQL::Test::Cluster;
```
AndersAstrand, line 5: While at it, maybe we should use the exact same preamble as is documented in `src/test/perl/README`?

```
    use strict;
    use warnings FATAL => 'all';
    use PostgreSQL::Test::Cluster;
    use PostgreSQL::Test::Utils;
    use Test::More;
``` 

The difference is the `use warnings` line. I have no idea how the two differs in actual behavior though.

t/basic.pl
```diff
@@ -34,68 +32,61 @@
...
 
-PGTDE::append_to_result_file(
-	'TABLEFILE FOUND: ' . (-f $tablefile ? 'yes' : 'no'));
-
-my $strings = 'CONTAINS FOO (should be empty): ';
-$strings .= `strings $tablefile | grep foo`;
-PGTDE::append_to_result_file($strings);
-
+ok(-f $tablefile, qq("$tablefile" exists));
+is(`strings $tablefile | grep foo`, '', 'should not find plain text');
```
AndersAstrand: I think we can go straight to something like
```
my $tablefilecontents = slurp_file($tablefile);

unlike($tablefilecontents, qr/foo/,
	'table file does not contain plaintext data');
```

instead here. `slurp_file()` from `PostgreSQL::Test::Utils` will error if the file doesn't exist :shrug:. It feels weird for our test to assert that postgres doesn't lie to us about what the file is called and calling out to `strings` feels even weirder 😆

t/basic.pl
jeltz: My plan was to do that as a separate commit which fixes all of them after researching the difference.

t/basic.pl
jeltz: My worry was that I did not want to output binary data but will give it a shot and see.

jeltz: And, yes, I should not use `ok()` if I want to fail if the file does not exist that is a `BAIL_OUT()`.

AndersAstrand: > My worry was that I did not want to output binary data but will give it a shot and see.

a yes. ok() with a regex comparison then perhaps :)

jeltz: Fixed!

t/tde_heap_aes_256.pl
```diff
@@ -169,23 +135,62 @@ sub verify_table
...
-done_testing();
+	my $tablefile =
+	  $node->data_dir . '/'
+	  . $node->safe_psql('postgres',
+		"SELECT pg_relation_filepath('$table');");
+
+	$stdout =
+	  $node->safe_psql('postgres', "SELECT * FROM $table ORDER BY id;");
+
+	if ($table eq 'test_enc0')
```
AndersAstrand: This is horrible. Why can't the assertions be inlined where they're used instead of this if thingy?

jeltz: I will give it a shot. I tried to keep the changes small but maybe this was too small. This horror was jsut something exposed here.

jeltz: Fixed.

t/basic.pl
```diff
@@ -34,68 +32,67 @@
 			(grantee IS NULL OR grantee = 0)
 		ORDER BY pg_proc.oid::regprocedure::text;
 	});
+is( $stdout,
+	"pg_tde_is_encrypted(regclass)\npg_tde_version()",
```
dutow, line 36: Lines with "\n" aren't too readable, shouldn't we write this differently?

t/basic.pl
```diff
@@ -34,68 +32,67 @@
...
-PGTDE::psql($node, 'postgres',
-	"SELECT pg_tde_set_key_using_database_key_provider('test-db-key', 'file-vault');"
-);
-
-PGTDE::psql($node, 'postgres',
-	'CREATE TABLE test_enc (id SERIAL, k VARCHAR(32), PRIMARY KEY (id)) USING tde_heap;'
-);
+$node->safe_psql(
+	'postgres', qq(
+	SELECT pg_tde_add_database_key_provider_file('file-vault', '$keydir/db.keys');
```
dutow, line 53: if we are already refactoring, maybe we should start renaming all the "file-vault" mentions? It's a quite confusing name.

t/basic.pl
```diff
@@ -34,68 +32,67 @@
...
-PGTDE::psql($node, 'postgres', 'SELECT * FROM test_enc ORDER BY id;');
+$stdout = $node->safe_psql('postgres', 'SELECT * FROM test_enc ORDER BY id;');
+is($stdout, "1|foobar\n2|barfoo", 'can read encrypted table');
 
-PGTDE::append_to_result_file("-- server restart");
 $node->restart;
 
-PGTDE::psql($node, 'postgres', 'SELECT * FROM test_enc ORDER BY id;');
+$stdout = $node->safe_psql('postgres', 'SELECT * FROM test_enc ORDER BY id;');
+is($stdout, "1|foobar\n2|barfoo", 'can read encrypted table');
```
dutow: can read encrypted table after restart?

t/tde_heap_aes_256.pl
```diff
@@ -2,190 +2,186 @@
...
-	'CREATE TABLE test_enc0 (id SERIAL, k VARCHAR(32), PRIMARY KEY (id)) USING tde_heap;'
-);
-
-PGTDE::psql($node, 'postgres',
-	"INSERT INTO test_enc0 (k) VALUES ('multitude'), ('multitudinous');");
-
-PGTDE::psql($node, 'postgres', 'SELECT * FROM test_enc0 ORDER BY id;');
+$node->safe_psql(
+	'postgres', qq(
+	CREATE EXTENSION pg_tde;
```
dutow, line 20: or maybe we could add a helper fort his, as we repeat it in many tests?

t/basic.pl
jeltz: Is this better?

t/basic.pl
jeltz: Any suggestion? otherwise I think it is maybe a job for a separate PR.

t/basic.pl
jeltz: Fixed.

t/tde_heap_aes_256.pl
jeltz: Not sure if it is worth it. I think the current code is pretty clean.

AndersAstrand (approved):
I love it!

t/basic.pl
```diff
@@ -34,68 +32,68 @@
 			(grantee IS NULL OR grantee = 0)
 		ORDER BY pg_proc.oid::regprocedure::text;
 	});
+is_deeply(
```
AndersAstrand: I don't think we need to bend over backwards to make this pretty. I think the raw string is fine for now and if we have tests where we need the assertion to look prettier in the future we can implement a helper or something. This is still miles better than just comparing the whole output as a big blob. Let's not get lost in the weeds :)

jeltz: @dutow What do you think? I think either is fine and I am not convinced this version with split even is less bad due to how ugly Perl is.

jeltz: Actually with syntax highlightning of \n which I have I kinda prefer the original.

dutow: yes this is not better, I was hoping we could use multi line strings or something simple

t/basic.pl
dutow: yes, maybe separate. "file.keyring" or something like that? "file-vault" always confuses me because we do have hashicorp vault support

t/basic.pl
jeltz: I will revert it for now and then maybe someone else or a future me can give it a new shot. I did some attempts with multiline strings too but they were not nice either, a lot due to perltidy.

t/basic.pl
jeltz: Yeah, I too dislike that name.

PR #591, opened 2026-05-27
Remove broken multiple_extensions.pl test

jeltz:
Since this test is disabled by default it broke a long time ago, and given that the idea of testing with multiple extensions as part of our TAP test is dubious I do not think it is worth fixing it.

The test has been broken without any complaints since at least since commit 36b81f18ddf14891f51e461974a7bcc4c690170e.


AndersAstrand (approved):
yeah, this belongs in the packaging tests, not here.

PR #592, opened 2026-05-27
Remove the Make builds

jeltz:
We kept the ability to build the project via make for some time to give more time for migrating from make to Meson, but now it is time to remove everything related to make in decrease the maintance costs when writing new code.

This makes it possible for us to remove almost everything from the .gitignore since build artefacts will end up in Meson's build directory only. The few remaining things in the .gitignore are of dubious value but let's remove them in a separate commit.


.gitignore
```diff
@@ -3,65 +3,10 @@
...
-objfiles.txt
-.deps/
-*.gcno
-*.gcda
-*.gcov
-*.gcov.out
-lcov*.info
-coverage/
-coverage-html-stamp
 *.vcproj
```
dutow: we can probably drop these 2? we don't have a visual studio project for pg_tde, this seems like a leftover from the repo split?

jeltz: It does. I can push a separate commit which removes them.

jeltz: Removed them.

dutow (approved):

PR #594, opened 2026-05-28
Make WAL and crash recovery TAP tests more idiomatic

jeltz:
Instead of trying to emulate pg_regress we should write idiomatic TAP tests. This makes them easier to work with and easier to understand for someone new to the project.

This exposed a probably bug in replication.pl where it does not seem to actually enable WAL encryption before generating the WAL.

Question: for reviewers: should I add a commit which fixes the bug or should I do it in a separate PR?


t/crash_recovery.pl
```diff
@@ -20,137 +21,115 @@
...
+	CREATE TABLE test_plain (x int PRIMARY KEY) USING heap;
+	INSERT INTO test_plain (x) VALUES (3), (4);
+
+	ALTER SYSTEM SET pg_tde.wal_encrypt = 'on';
+));
+
 $node->kill9;
 
-PGTDE::append_to_result_file("-- server start");
 PGTDE::poll_start($node);
```
AndersAstrand, line 48: This crash/restart cycle should have a comment as the others. Since "server is starting" is the actual assertion we do.

Maybe "server starts after crash when wal encryption was enabled but no wal generated yet" or something like that?

But I understand if that is seen as a pre-existing condition and out of scope of this PR. It's great that it's now jarring that we have an assertion without an explanation about what's being asserted though, something that was hidden in the .out files before.

t/crash_recovery.pl
```diff
@@ -20,137 +21,115 @@
...
 
-PGTDE::psql($node, 'postgres', "INSERT INTO test_enc (x) VALUES (7), (8);");
-PGTDE::append_to_result_file("-- kill -9");
+$node->safe_psql('postgres', 'INSERT INTO test_enc (x) VALUES (7), (8);');
+
 $node->kill9;
-PGTDE::append_to_result_file("-- change cipher to aes_256");
+
+# change cipher to aes_256
```
AndersAstrand: This useless comment didn't really have to survive any more than "-- kill -9" and the others though :D

t/crash_recovery.pl
```diff
@@ -20,137 +21,115 @@
...
-PGTDE::psql($node, 'postgres', "SELECT nextval('seq_unlogged');");
+$node->safe_psql('postgres',
+	'CREATE UNLOGGED SEQUENCE seq_unlogged OWNED BY test_enc.x;');
+
+$stdout =
+  $node->safe_psql('postgres', "SELECT pg_tde_is_encrypted('seq_unlogged');");
+is($stdout, 't', 'sequence is encrypted');
+
+$stdout = $node->safe_psql('postgres', "SELECT nextval('seq_unlogged');");
+is($stdout, '1', 'sequence has been incremented');
```
AndersAstrand: This is not nice at all. Asserting test setup is definitely not right.  Not sure how to fix it but I didn't set us up for success when I wrote it just leaning into the pg_regress style. We can fix it later however.

t/replication.pl
```diff
@@ -26,78 +27,67 @@
...
+	INSERT INTO test_plain (x) VALUES (3), (4);
+));
 
-PGTDE::append_to_result_file("-- At replica");
+$primary->wait_for_catchup('replica');
 
-PGTDE::psql($replica, 'postgres', "SELECT pg_tde_is_encrypted('test_enc');");
-PGTDE::psql($replica, 'postgres',
+$stdout =
+  $replica->safe_psql('postgres', "SELECT pg_tde_is_encrypted('test_enc');");
```
AndersAstrand, line 47: We really need to clean up all of these assertions on test setup. That `USING tde_heap` even works at all is not this test suite's job to assert. It's done in basic.pl so here it should just be assumed to work unless we're testing something that would possibly break it here.

t/replication.pl
```diff
@@ -26,78 +27,67 @@
...
-	"INSERT INTO test_enc2 (x) VALUES (1), (2);");
-
-PGTDE::psql($primary, 'postgres',
-	"ALTER SYSTEM SET pg_tde.wal_encrypt = 'on';");
+is($stdout, 'f', 'test_plain_pkey is not encrypted');
+$stdout =
+  $replica->safe_psql('postgres', 'SELECT * FROM test_plain ORDER BY x;');
+is($stdout, "3\n4", 'can read from test_plain');
+
+# check primary crash with WAL encryption
```
AndersAstrand, line 66: I am confused here. This first crash/restart cycle seems to assert that the primary starts when wal_encryption is enabled but no WAL is generated before the crash. Right?

AndersAstrand (approved):
So a lot of these test have very unclear setup/execute/assert cycles, something that writing them in this way makes obvious which is a clear win. Not sure how much of this to fix in this PR. Maybe none and just let it be as bad as it was before but visible.

Not sure about the TODO comment though?

t/replication.pl
jeltz: No, this actually asserts something real. Read it again.

jeltz: But it should explain it better.

t/replication.pl
jeltz: This is why I added a TODO, the test does not seem to test anything useful.

t/crash_recovery.pl
jeltz: Fixed.

t/crash_recovery.pl
jeltz: Fixed.

t/crash_recovery.pl
jeltz: That one was easy to remove but I am a bit skeptical about removing the sanity check for that the sequence is actually encrypted.

PR #595, opened 2026-06-01
PG-2125: Switching kmip support to the new C++ kmip client library

dutow:
This commit updates the submodule pointer for kmip and reworks our code to use a new C++ api instead of the old and buggy C api.


ci_scripts/run-pgindent.sh
```diff
@@ -19,4 +19,4 @@ cd "$SCRIPT_DIR/.."
 export PATH=$SCRIPT_DIR/../../postgres/src/tools/pgindent/:$INSTALL_DIR/bin/:$PATH
 
 # Check pg_tde with the fresh list extraxted from the object file
-pgindent --typedefs=typedefs.list --excludes=<(echo "src/libkmip") "$@" .
+pgindent --typedefs=typedefs.list "$@" src
```
dAdAbird: Wouldn't it skip `fetools` now?

dAdAbird (commented):
LGTM overall, I'm just not sure about the pgindent scope change

.github/workflows/sanitizers.yml
```diff
@@ -12,11 +12,11 @@ on:
 env:
   pg_version: 18
   CC: clang
+  CXX: clang++
```
jeltz, line 15: Isn't this in the wrong commit?

.github/workflows/coverage.yml
```diff
@@ -50,8 +50,8 @@ jobs:
 
       - name: Process coverage
         run: |
-          geninfo -o coverage.info --rc lcov_branch_coverage=1 -i .
-          geninfo -o coverage.info --rc lcov_branch_coverage=1 .
+          geninfo -o coverage.info --rc lcov_branch_coverage=1 --ignore-errors mismatch -i .
+          geninfo -o coverage.info --rc lcov_branch_coverage=1 --ignore-errors mismatch .
```
jeltz, line 55: Why was ignore errors necessary? The commit message does not say why.

ci_scripts/run-clang-format.sh
```diff
@@ -0,0 +1,33 @@
...
+
+SCRIPT_DIR="$(cd -- "$(dirname "$0")" >/dev/null 2>&1; pwd -P)"
+cd "$SCRIPT_DIR/.."
+
+CLANG_FORMAT="${CLANG_FORMAT:-clang-format}"
+
+FILES=$(git ls-files 'src/**/*.cpp')
+
+if test -z "$FILES"; then
+	echo "no C++ sources found"
```
jeltz: Either skip this message or write to stderr.

ci_scripts/run-clang-format.sh
```diff
@@ -0,0 +1,33 @@
...
+# C sources and headers are handled by pgindent instead; see run-pgindent.sh.
+
+set -e
+
+SCRIPT_DIR="$(cd -- "$(dirname "$0")" >/dev/null 2>&1; pwd -P)"
+cd "$SCRIPT_DIR/.."
+
+CLANG_FORMAT="${CLANG_FORMAT:-clang-format}"
+
+FILES=$(git ls-files 'src/**/*.cpp')
```
jeltz: This code does not support file names with spaces in them, but not sure if we care.

meson.build
```diff
@@ -92,16 +92,21 @@ install_data(
...
+})
+kmip_cmake_opts.set_install(false)
+# libkmip is a vendored external dependency. Don't impose our -Dwerror=true on
+# it: its own warnings (e.g. -Woverloaded-virtual on newer compilers) must not
+# fail our build. -Wno-error is appended after meson's injected -Werror.
+kmip_cmake_opts.append_compile_args('c', '-Wno-error')
+kmip_cmake_opts.append_compile_args('cpp', '-Wno-error')
+kmip_subproject = cmake.subproject('libkmip', options: kmip_cmake_opts)
+kmipclient_dep = kmip_subproject.dependency('kmipclient')
+kmipcore_dep = kmip_subproject.dependency('kmipcore')
```
jeltz: We do not use `_dep` suffixes for depenenceis. Maybe we should but if so we should change all.

src/keyring/keyring_kmip.cpp
```diff
@@ -0,0 +1,196 @@
...
+	char errbuf[256];
+	bool ok = true;
+
+	try
+	{
+		body();
+	}
+	catch (const std::exception &e)
+	{
+		snprintf(errbuf, sizeof(errbuf), "%s", e.what());
```
jeltz: Shouldn't we use strlcpy?

src/keyring/keyring_kmip.cpp
```diff
@@ -0,0 +1,196 @@
...
+	         });
+}
+
+static KeyInfo *
+kmip_get_key(GenericKeyring *keyring, const char *key_name,
+             KeyringReturnCode *return_code)
+{
+	KmipKeyring *kmip_keyring = (KmipKeyring *)keyring;
+	unsigned char keydata[MAX_KEY_DATA_SIZE];
+	int keylen = 0;
```
jeltz, line 101: This should probably not be initialized unless it is necessary to silence the compiler,

src/keyring/keyring_kmip.cpp
```diff
@@ -0,0 +1,196 @@
...
+	{
+		ereport(WARNING,
+		        errmsg("keyring provider returned invalid key size"));
+		*return_code = KEYRING_CODE_INVALID_KEY;
+		return NULL;
+	}
+
+	key = palloc_object(KeyInfo);
+	memset(key->name, 0, sizeof(key->name));
+	memcpy(key->name, key_name, strnlen(key_name, sizeof(key->name) - 1));
```
jeltz: I suspect this is old code but isn't this all a more convoluted way to do:

```
strncpy(key->name, key_name, sizeof(key->name));
key->name[sizeof(key->name) - 1] = '\0';
```

or

```
memset(key->name, 0, sizeof(key->name));
strlcpy(key->name, key_name, sizeof(key->name));
```

src/keyring/keyring_kmip.cpp
```diff
@@ -0,0 +1,196 @@
...
+}
+
+static const TDEKeyringRoutine keyringKmipRoutine = {
+    .keyring_get_key = kmip_get_key,
+    .keyring_store_key = kmip_set_key,
+    .keyring_validate = kmip_validate,
+};
+
+void
+InstallKmipKeyring(void)
```
jeltz: Since this is C++ should there actually be a `void` here? Or is that a thing one does in `extern "C"` blocks?

src/keyring/keyring_kmip.cpp
```diff
@@ -0,0 +1,196 @@
...
+kmip_connect(KmipKeyring *kmip_keyring)
+{
+	return kmipclient::Kmip(kmip_keyring->kmip_host, kmip_keyring->kmip_port,
+	                        kmip_keyring->kmip_cert_path,
+	                        kmip_keyring->kmip_key_path,
+	                        kmip_keyring->kmip_ca_path, KMIP_TIMEOUT_MS);
+}
+} /* namespace */
+
+extern "C"
```
jeltz: I wonder if it would not be nicer to put `extern "C"` per function instead of as a block. especially with the PG coding style.

jeltz (changes requested):

ci_scripts/run-pgindent.sh
dutow: fixed

.github/workflows/sanitizers.yml
dutow: fixed

.github/workflows/coverage.yml
dutow: explained in commit message, otherwise we'll get errors because of the C++ parts

ci_scripts/run-clang-format.sh
dutow: fixed

ci_scripts/run-clang-format.sh
dutow: fixed

meson.build
dutow: fixed

src/keyring/keyring_kmip.cpp
dutow: fixed

src/keyring/keyring_kmip.cpp
dutow: fixed

src/keyring/keyring_kmip.cpp
dutow: fixed and used palloc0_object instead

src/keyring/keyring_kmip.cpp
dutow: no, fixed

src/keyring/keyring_kmip.cpp
dutow: we don't even need extern C as we only expose function pointers. I removed it completely.

jeltz (approved):
Approved, but see my comment on Slack about way forward.

PR #596, opened 2026-06-01
PG-2399 Implement pg_checksums for encrypted tables

jeltz:
There was nothing fundamental in our design which prevented  `pg_tde_checksunms` from validating or enabling checksums on encrypted tables, so do that instead of just skipping them.

Also enable two previously skipped PostgreSQL tests, one which now passes and the other which always seems to have passed.

### DIFF
--- a/ci_scripts/suppressions/lsan.supp
+++ b/ci_scripts/suppressions/lsan.supp
@@ -88,6 +88,3 @@ leak:HandleSlashCmds
 
 #pg_restore
 leak:RestoreArchive
-
-#pg_tde_checksums
-leak:get_principal_key_from_keyring
--- a/ci_scripts/suppressions/lsan.supp
+++ b/ci_scripts/suppressions/lsan.supp
@@ -88,3 +88,6 @@ leak:HandleSlashCmds
 
 #pg_restore
 leak:RestoreArchive
+
+#pg_tde_checksums
+leak:get_principal_key_from_keyring
--- a/fetools/pg17/pg_checksums/pg_checksums.c
+++ b/fetools/pg17/pg_checksums/pg_checksums.c
@@ -197,7 +197,7 @@ scan_file(const char *fn, Oid spcOid, Oid dbOid, RelFileNumber relNumber, ForkNu
 	int			flags;
 	int64		blocks_written_in_file = 0;
 	RelFileLocator locator = {.spcOid = spcOid,.dbOid = dbOid,.relNumber = relNumber};
-	InternalKey *key;
+	InternalKey *key = NULL;
 
 	Assert(mode == PG_MODE_ENABLE ||
 		   mode == PG_MODE_CHECK);
@@ -210,7 +210,9 @@ scan_file(const char *fn, Oid spcOid, Oid dbOid, RelFileNumber relNumber, ForkNu
 
 	files_scanned++;
 
-	key = pg_tde_get_smgr_key(locator);
+	/* Unknown fork type so assume it is not encrypted */
+	if (forknum != InvalidForkNumber)
+		key = pg_tde_get_smgr_key(locator);
 
 	for (blockno = 0;; blockno++)
 	{
--- a/fetools/pg17/pg_checksums/pg_checksums.c
+++ b/fetools/pg17/pg_checksums/pg_checksums.c
@@ -131,14 +131,6 @@ pg_tde_init(const char *datadir)
 	pg_tde_fe_init(tdedir);
 }
 
-static bool
-is_pg_tde_encypted(Oid spcOid, Oid dbOid, RelFileNumber relNumber)
-{
-	RelFileLocator locator = {.spcOid = spcOid,.dbOid = dbOid,.relNumber = relNumber};
-
-	return pg_tde_has_smgr_key(locator);
-}
-
 /*
  * Report current progress status.  Parts borrowed from
  * src/bin/pg_basebackup/pg_basebackup.c.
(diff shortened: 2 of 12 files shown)


Andriciuc (approved):
Docs look good, thank you for updating them!

fetools/pg17/pg_checksums/pg_checksums.c
```diff
@@ -396,21 +404,17 @@ scan_directory(const char *basedir, const char *subdir, Oid tablespace, bool siz
 				/* filenode not to be included */
 				continue;
 
-			dirsize += st.st_size;
+			if (forkpath != NULL)
+				forknum = forkname_to_number(forkpath);
```
dAdAbird, line 410: Should we check for InvalidForkNumber and throw an error/warning if it was returned?
The same in pg18/

t/pg_tde_checksums.pl
```diff
@@ -0,0 +1,56 @@
...
+
+my $stderr;
+
+my $keydir = PostgreSQL::Test::Utils::tempdir;
+
+my $node = PostgreSQL::Test::Cluster->new('main');
+$node->init(no_data_checksums => 1);
+$node->append_conf('postgresql.conf', "shared_preload_libraries = 'pg_tde'");
+$node->start;
+$node->safe_psql(
```
dutow, line 15: I still wonder if a basic "setup an encrypted cluster" helper would help us, as we keep repeating the same thing in tests.

dutow (approved):
The changes look good - but maybe a better long term solution would be to change our solution so that checksums work without encryption knowledge, by ignoring recalculating the checksum field after encryption/decryption. That way external tools, such as pgbackrest, would also be able to verify checksums.

ci_scripts/suppressions/lsan.supp
```diff
@@ -88,3 +88,6 @@ leak:HandleSlashCmds
 
 #pg_restore
 leak:RestoreArchive
+
+#pg_tde_checksums
+leak:get_principal_key_from_keyring
```
dAdAbird: Should/can we fix the leak there? I understand in might be only FE related and doesn't meter here. But this func is used by the backend as well and skipping it in sanitizers might cause some issues in the future (refactoring etc.)

dAdAbird (commented):
Looks good to me in general. Just a few comments/questoins

t/pg_tde_checksums.pl
jeltz: Maybe, if you feel like it give it a shot.

fetools/pg17/pg_checksums/pg_checksums.c
jeltz: Nah, I think we should just assume it is not encrypted if it is some unknown type. Does that make sense?

ci_scripts/suppressions/lsan.supp
jeltz: Yeah, this is for sure bad. But should we try to cache principal keys or is it ok to fetch them once for every table?

jeltz:
The question if we should fix checksums is tricky and I think bigger than this issue. I am not sure it is even safe to not encrypt the checksums.

ci_scripts/suppressions/lsan.supp
dutow: I ignored this because we are using palloc and caching on the backend, so we should be relatively safe there. Depending on how much complexity it adds to the fronnend, it might not be worth fixing.

fetools/pg17/pg_checksums/pg_checksums.c
dAdAbird: Yeah, makes sense

ci_scripts/suppressions/lsan.supp
dAdAbird: Yes, it uses palloc, but I more concerned about if during some change/refactoring of get_principal_key_from_keyring() or **_any other function or code that invokes it_** - as far as I remember suppressions matched against the stack trace -  we introduce some leak, we'll never find it out because of the suppression. As for me, it is better to suppress-match some function in the FE tool (even with wider scope) or the tool itself (it's a frontend who cares) then a function that invoked by backend. 

> Yeah, this is for sure bad. But should we try to cache principal keys or is it ok to fetch them once for every table?

Btw, I one of the pg_rewind related changes I just [introduced `clean_fe_server_principal_key_cache()`](https://github.com/percona/pg_tde/commit/39f9b53faf9f7eb198b2f4d91fb1d999bd5f5f9c#diff-fc8109ce322645740f720cb5a9101e7aef73c26a54354df2abc93d60d5007111R959-R965)
It's kinda necessity there due to the need of switching between different pg_tde dirs, hence a need to clean up caches. 

Idk, maybe leave some TODO somewhere here and get back to it when/if we merge those changes of mine. 
But I'd still prefer to put into supp some pg_checksums related function rather then backend

jeltz: I think the proper solution is the add a `pfree()` call to `pg_tde_get_smgr_key()` when `FRONTEND` is defined, like I have now done.

dAdAbird: Yes, looks good. So we can remove this entry from lsan.supp‎ now?

dAdAbird (approved):

ci_scripts/suppressions/lsan.supp
jeltz: Yes, I did!

PR #599, opened 2026-06-03
pg_rewind: ensure WAL segments restored from archive

dAdAbird:
Before this change, all segments that were restored from the archive by extractPageMap flew under the radar as it run after the rewind built map files on the destination. Hence, all these segments are just neglected by the later actions. This change makes all restored data from the active segments to be re-encrypted later on.

Fix PG-2397
For PG-2407

### DIFF
--- a/fetools/pg18/pg_rewind/filemap.h
+++ b/fetools/pg18/pg_rewind/filemap.h
@@ -116,7 +116,11 @@ extern void print_filemap(filemap_t *filemap);
 
 extern void keepwal_init(void);
 extern void keepwal_add_entry(const char *path);
+extern bool keepwal_entry_exists(const char *path);
 
 extern bool path_rlocator(const char *path, RelFileLocator *rlocator, unsigned int *segNo);
+
+extern void ensure_tde_archive_wal(void);
+extern void archive_wal_segments_add_entry(const char *path);
 
 #endif							/* FILEMAP_H */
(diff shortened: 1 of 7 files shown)


fetools/pg18/pg_rewind/parsexlog.c
```diff
@@ -349,8 +349,14 @@ SimpleXLogPageRead(XLogReaderState *xlogreader, XLogRecPtr targetPagePtr,
 			if (xlogreadfd < 0)
 				return -1;
 			else
+			{
+				char		xlogdirfname[MAXFNAMELEN];
```
dutow: Should be MAXPGPATH?

fetools/pg18/pg_rewind/filemap.h
```diff
@@ -116,7 +116,11 @@ extern void print_filemap(filemap_t *filemap);
 
 extern void keepwal_init(void);
 extern void keepwal_add_entry(const char *path);
+extern bool keepwal_entry_exists(const char *path);
 
 extern bool path_rlocator(const char *path, RelFileLocator *rlocator, unsigned int *segNo);
 
+extern void ensure_tde_archive_wal();
```
dutow: Should be ensure_tde_archive_wal(void)?

dutow (changes requested):

dutow (approved):

PR #600, opened 2026-06-04
Remove the last uses of test helpers which emulated pg_regress

jeltz:
These are the last tests which used the test helpers so we can remove the helpers too.


dAdAbird (commented):
I think it should be "uses" not "users" in the commit message. LGTM otherwise

dAdAbird (approved):

PR #601, opened 2026-06-04
Remove stream_rep.pl which just duplicates test in PostgreSQL

jeltz:
Since we now run the test which this is based on, `src/test/recovery/t/001_stream_rep.pl`, as part of our CI when we run the PostgreSQL test suite with `pg_tde` enabled there is no reason to keep it around. The test had no additional modification outside enabling `pg_tde`.


dAdAbird (commented):
> The test had no additional modification outside enabling pg_tde.

This test also uses PGTD::backup() which does copy of pg_tde dir prior to basebackup and runs pg_tde_basebackup with -E (wal_encryption). I'm not sure though how useful this test is if the upstream one passes anyway. So maybe we can remove it. But we definitely have to make a proper replication test for tde

jeltz:
We do that too in the PSP tests.

```
sub backup
{
	my ($self, $backup_name, %params) = @_;
	my $backup_dir = $self->backup_dir . '/' . $backup_name;

	mkdir $backup_dir or die "mkdir($backup_dir) failed: $!";

	if ($tde_mode_wal)
	{
		# TODO: More thorough checking for options incompatible with --encrypt-wal
		$params{backup_options} = [] unless defined $params{backup_options};
		unless (
			List::Util::any { $_ eq '-Ft' or $_ eq '-Xnone' }
			@{ $params{backup_options} })
		{
			PostgreSQL::Test::Utils::system_log('cp', '-R', '-P', '-p',
				$self->pg_tde_dir, $backup_dir . '/pg_tde',);

			push @{ $params{backup_options} }, '--encrypt-wal';
		}
	}

	$self->SUPER::backup($backup_name, %params);
}
```

dAdAbird:
Ah, great

dAdAbird (approved):

PR #602, opened 2026-06-04
Fix re-encryption of segmented relations

dAdAbird:
If the relation is segmented, there is no guarantee that the pg_rewind
processes all segments in the sequence. So the approach of dumping the
relation key into the source when we switch to the next file, or even
with rlocator, won't work. It is common to have the next sequence of
file processing:
base/16384/16443
base/16384/16440
base/16384/16443.1
So by the time we get to the 16443.1 segment, the keys for 16443 are
rewritten, and it can't properly decrypt it.

To fix this, instead of writing keys on the file switch, we add them
to the hashmap and process them in the end, after all the files.

It also fixes segmented forks (VM/FSM).
A TAP test for segmented forks would be problematic as it'd require a relation of ~3.5Tb size

For PG-2407

### DIFF
--- a/fetools/pg18/pg_rewind/tde_ops.c
+++ b/fetools/pg18/pg_rewind/tde_ops.c
@@ -16,6 +16,7 @@
 #include "access/pg_tde_xlog_keys.h"
 #include "access/pg_tde_xlog_smgr.h"
 #include "common/pg_tde_utils.h"
+#include "common/hashfn.h"
 #include "pg_tde.h"
 
 static void copy_dir(const char *src, const char *dst);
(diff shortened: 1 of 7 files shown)


fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -31,105 +32,193 @@ typedef struct
...
-		 * Server can recover from wrecked VM/FSM, hence only warnings here
-		 * and in the rest of the function
-		 */
-		pg_log_warning("could not open fork file \"%s\": %m", srcpath);
-		pg_log_warning_hint("%s", warning_hint);
-		return;
+		entry->rloc = rloc;
+		entry->source_key = source_key;
+		entry->target_key = target_key;
 	}
```
dutow, line 117: won't we possibly leak source/target key if(found), e.g. on multiple segments? And a related question, should we have at least an assertion in release builds that if we found it, it's the same pair?

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -31,105 +32,193 @@ typedef struct
...
-	off_t		offset;
-	RelPathStr	rp = relpathperm(current_tde_file.rlocator, fork);
-	static const char *const warning_hint = "Skipping the file, as the server can start and rebuild the broken VM/FSM file.";
+	flushkey = flushkey_create(FLUSHKEY_INIT_SIZE, NULL);
+}
 
-	if (dry_run)
-		return;
+/*
+ * Add used a key that was used for relation re-encrion to the hash. We will
```
dutow: re-encryption

fetools/pg18/pg_rewind/tde_ops.c
dAdAbird: Good catch. Fixed. I decided to trade extra allocations for code clarity. So we just do copies here, and callers may free keys.

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -44,105 +45,209 @@ typedef struct
...
+		if (!dry_run)
+		{
+			reencrypt_rel(entry->rloc, FSM_FORKNUM, entry->source_key, entry->target_key);
+			reencrypt_rel(entry->rloc, VISIBILITYMAP_FORKNUM, entry->source_key, entry->target_key);
+			pg_tde_save_smgr_key(entry->rloc, entry->target_key, true);
 		}
-		offset += read_len;
+
+		pg_free(entry->source_key);
+		pg_free(entry->target_key);
```
dutow, line 161: shouldn't we also null these?

fetools/pg18/pg_rewind/tde_ops.c
```diff
@@ -245,15 +355,25 @@ ensure_tde_keys(const char *relpath)
...
-
-	if (!path_rlocator(relpath, &rlocator, &segNo))
+	/* the same relation, nothing to do */
+	if (RelFileLocatorEquals(rlocator, current_tde_file.rlocator))
 		return;
 
+	/*
+	 * This is a new relation. Save keys for later processing.
+	 */
+	if (current_tde_file.source_key != NULL)
```
dutow, line 374: Won't this spam pg_tde_get_smgr_key requests for unencrypted tables? Since we never cache that the relation has no key. I think it was the same before, so this isn't a new issue.

dutow (approved):
I added two more comments, but these seem optional to me.

fetools/pg18/pg_rewind/tde_ops.c
dAdAbird: This code is reachable only if it is a new rlocator, hence a new file, and blocks of a file are processed sequentially. So there will be some extra checks when switching to a new segment of unencrypted relation, but I think we can ignore that for now.

fetools/pg18/pg_rewind/tde_ops.c
dAdAbird: I've added `flushkey_destroy(flushkey);` after the loop to ensure everything is cleaned up and the table won't be used.

PR #604, opened 2026-06-04
PG-2398 Convert default principal key test to TAP

jeltz:
Since the `pg_buffercache_evict()` function does not exist in PostgreSQL 16 we cannot test reading from disk in a pg_regress test there which means we need to implement it as a TAP test with restarts the cluster.

And in my opinion this makes the test much easier to understand anyway.


t/default_principal_key.pl
```diff
@@ -0,0 +1,143 @@
...
+	'SELECT id, name FROM pg_tde_list_all_global_key_providers();');
+is($stdout, '-1|file-provider',
+	'key provider is still there after failed delete');
+
+# Test key roatation
+
+$node->safe_psql('postgres', 'CREATE DATABASE other');
+$node->safe_psql('other', 'CREATE EXTENSION pg_tde;');
+
+# postgres
```
dutow: maybe "postgres database" or "database: postgres"? At first I didn't understand this commen here

dutow (approved):

t/default_principal_key.pl
jeltz: Fixed!

PR #605, opened 2026-06-05
Fix dry_run issues

dAdAbird:
It fixes ensure_tde_wal_seg() and reencrypt_fork(), but doesn't touch flush_current_tde_rel_key() as this is gone in another PR and I'll fix it there

!!! Needs a TAP test !!!

For PG-2407


jeltz:
You might as well merge this since this is not a complete solution anyway due to the thing you wrote yourself.

jeltz (approved):
This is hard to review on its own so just merge it into the working branch and the commits where it belongs.

PR #607, opened 2026-06-07
Remove 2PC test which is mostly copied from PostgreSQL

jeltz:
Our test for two-phase commit are almost directly copied from the PostgreSQL test `src/test/recovery/t/009_twophase.pl` but with the only real addition being that we assert that the table is encrypted. It is unclear to me this small thing is worth the maintenance burden so this removes the test.

@dutow This is a case where I am not sure, maybe we should add a simple 2PC test.


dutow (approved):
> we assert that the table is encrypted.

Maybe we can do the same I did with stormweaver, where we verify the entropy of encrypted files at the end of the testrun? But somehow integrated into the pg test framework

jeltz:
@dutow Yeah that could make sense.

PR #608, opened 2026-06-08
PG-2439 Build with all LDFLAGS from PostgreSQL to fix RPATH builds

jeltz:
When PostgreSQL was built with RPATH that was not honored when we built `pg_tde` which could mean that the backend may load one version of `libpq` while our `pg_tde_*` frontend tools could load another and that can cause issues at least with the `--write-recovery-conf` argument to pg_rewind when `GenerateRecoveryConfig()` calls into `libpq` and gets options not supported by the version of `libpq` used by the backend.

To fix this we just blindly pass along ass `LDFLAGS` invluding (`LDFLAGS_EX` and `LDFLAGS_SL`) which should include all the options necessary for an RPATH build. This should also be good in general since the linker flags could potentially contain other important options.

The Meson code for this sadly ends up quite ugly due to PostgreSQL not supporting `pkgconfig` for extensions and us depending on a Meson version which does not support `splitlines()` yet.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,6 +7,7 @@ project('pg_tde',
 fs = import('fs')
 
 pg_config = find_program(get_option('pg_config'))
+shsplit = find_program(files('shsplit'))
 
 bindir = run_command(pg_config, '--bindir', check: true).stdout().strip()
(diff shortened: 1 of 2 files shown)


rjd15372 (approved):
lgtm

PR #613, opened 2026-06-10
Hide keyring debug messages behind KEYRING_DEBUG

dAdAbird:
Those messages just contaminate frontend tools' debug input without adding value

### DIFF
--- a/src/catalog/tde_keyring.c
+++ b/src/catalog/tde_keyring.c
@@ -773,9 +773,6 @@ scan_key_provider_file(ProviderScanType scanType, void *scanKey, Oid dbOid)
 			continue;
 		}
 
-		ereport(DEBUG2,
-				errmsg("read key provider ID=%d %s", provider.provider_id, provider.provider_name));
-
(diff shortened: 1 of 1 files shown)


src/catalog/tde_keyring.c
```diff
@@ -773,8 +773,10 @@ scan_key_provider_file(ProviderScanType scanType, void *scanKey, Oid dbOid)
 			continue;
 		}
 
+#ifdef KEYRING_DEBUG
```
jeltz: I think we should just remove this log statement.

jeltz (changes requested):
I think one of the changes look good but that one just should be removed.

src/catalog/tde_keyring.c
dAdAbird: done

jeltz (approved):

jeltz:
I think we could merge this into `main`.

dAdAbird:
I merged it into the rewind branch since it is much more needed there now. But eventually it gets into main, once we merge the rewind branch.

PR #615, opened 2026-06-11
pg_tde_rewind code quality improvements

dutow:
Pure refactors, no behaviour change.

  * pg_tde_set_data_dir: drop redundant memsets before strlcpy/snprintf
  * enc_tde: name the 16-byte IV constant TDE_BLOCK_IV_LEN
  * enc_tde: rename relKey -> rel_key in the new public block helpers
  * tde_ops copy_dir: simplify loop variable names
  * archive tools: extract shared tdedir derivation helper
  * tde_ops reencrypt_fork: use one O_RDWR fd + pread/pwrite
  * local_source: inline single-use directory_exists helper


dAdAbird (approved):
LGTM
Thanks!

PR #617, opened 2026-06-11
pg_rewind: honor TMPDIR for temporary pg_tde working directory

dutow:
The previous code hardcoded /tmp as the parent of the throwaway TDE key directory created via mkdtemp(). On hosts where /tmp is on a size-limited tmpfs, read-only, or simply absent, pg_tde_rewind would fail with an opaque "could not create temporary directory" error. Mirror PostgreSQL's convention of honouring $TMPDIR with a /tmp fallback.

While here, guard destroy_tde_tmp_dir (registered via atexit) so that an early abort before init_tde does not pass an empty path to rmtree.


dAdAbird (approved):
Btw, we do the same hardcoded tmp also in `restore_encrypt`, `archive_decrypt` and `pg_tde_upgrade`

PR #620, opened 2026-06-15
Clean up and standardize the beginnings of TAP files

jeltz:
The `use` directives of our TAP files were all over the place. Some fixes:

- Sort the `use` directives alphabetically
- Remove `use lib 't'`
- Remove shebangs
- Always use `use warnings FATAL => 'all'`

### DIFF
--- a/t/kmip.pl
+++ b/t/kmip.pl
@@ -3,7 +3,6 @@
 use strict;
 use warnings;
 use Test::More;
-use lib 't';
 use CosmianKms;
 use PostgreSQL::Test::Utils;
 use PostgreSQL::Test::Cluster;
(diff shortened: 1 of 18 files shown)


t/kmip.pl
```diff
@@ -1,14 +1,11 @@
...
-use Test::More;
-use lib 't';
+use warnings FATAL => 'all';
 use CosmianKms;
-use PostgreSQL::Test::Utils;
-use PostgreSQL::Test::Cluster;
 use IO::Socket::INET;
 use POSIX qw(:sys_wait_h);
+use Test::More;
+use PostgreSQL::Test::Utils;
```
dutow: nitpick: this isn't alphabetically sorted

dutow (approved):

t/kmip.pl
jeltz: Thanks!

PR #622, opened 2026-06-15
PG-2472: Fix pg_tde with EXEC_BACKEND

dutow:
We had two separate issues caused by reliance on fork behavior, fixed in two separate commits.

This commit doesn't add an EXEC_BACKEND CI yet because we need the PSP side changes merged first before that would work.


dAdAbird (commented):
Changes LGTM. But I have a question:
Since EXEC_BACKEND children don't inherit statics. Should we also take care of the [WAL write key](https://github.com/percona/pg_tde/blob/1dacc110831e6865cfb2435c328792620c176fef/src/access/pg_tde_xlog_smgr.c#L48)? We generate a new key and set it there when postmaster starts, and then backends use that key to encrypt WAL.

src/catalog/tde_principal_key.c
```diff
@@ -169,11 +169,17 @@ PrincipalKeyShmemInit(void)
...
-
 		dshash_detach(dsh);
 	}
+
+	/*
+	 * Set the local pointer in both branches. On fork-based platforms the
+	 * postmaster sets it before forking and children inherit it, but under
+	 * Windows EXEC_BACKEND each child starts with a NULL static, so every
+	 * process must install the pointer here.
+	 */
```
jeltz: This comment feels a bit unnecessary. The proper solution for this is CI, which we can add in a separate PR.

src/catalog/tde_keyring.c
```diff
@@ -91,19 +93,48 @@ static void check_provider_record(KeyringProviderRecord *provider_record);
...
+ * shared memory.
+ *
+ * GetNamedLWLockTranche() reads NamedLWLockTrancheRequestArray, which is a
+ * per-process static populated only by RequestNamedLWLockTranche() inside the
+ * shmem_request_hook. On fork-based platforms child processes inherit that
+ * static, so calling GetNamedLWLockTranche() in each shmem_startup_hook works.
+ * Under Windows EXEC_BACKEND the static is empty in children and calling it
+ * dereferences NULL. Store the pointer in shared memory once (when the
+ * postmaster first creates the shmem) so every process can find it.
+ */
```
jeltz, line 100: This looks like an accidental leftover from the LLM working. If we feel the need to comment about our previous missunderstanding of `GetNamedLWLockTranche()` we should do it in a more concise way.

jeltz (changes requested):
@dAdAbird's comment about the WAL key is interesting. Because it does seem like our code works with EXEC_BACKEND and encrypted WAL, hmmm.

src/catalog/tde_keyring.c
dutow: eh, yes, I thought I deleted it but looks like I forgot to.

src/catalog/tde_principal_key.c
dutow: This one I left intentionally here as a reminder why these can't be in the if above, but maybe mentioning it in the commit message is enough.

jeltz: I don't care strongly either way, for me commit message mention is good enough. I always look in the history if I wonder about something. Maybe see if you can tirm it down a bit if you want to keep it.

dAdAbird:
> @dAdAbird's comment about the WAL key is interesting. Because it does seem like our code works with EXEC_BACKEND and encrypted WAL, hmmm.

Does it actually encrypt WAL? tdeheap_xlog_seg_write() probably [falls to the regular pg_pwrite()](https://github.com/percona/pg_tde/blob/1dacc110831e6865cfb2435c328792620c176fef/src/access/pg_tde_xlog_smgr.c#L425-L436).

dutow:
The WAL find was a good catch, previously I didn't get CI working on windows, and I didn't try wal encryption manually.

I also added an exec backend CI to the PR now, as all other related PRs are merged.

ci_scripts/build.sh
```diff
@@ -31,9 +31,15 @@ case "$1" in
         ARGS+=" -Dc_link_args=['-fsanitize=address','-fsanitize=undefined']"
         ;;
 
+    exec-backend)
+        echo "Building with exec-backend option"
+        BUILD_TYPE=debug
```
jeltz: Should we use `debugoptimize`? I feel that is a tradeoff between build time vs time to execute tests.

ci_scripts/build.sh
```diff
@@ -31,9 +31,15 @@ case "$1" in
...
+        ARGS+=" -Dc_args=['-DEXEC_BACKEND']"
```
jeltz: This does not look to be strictly necessary right?

src/access/pg_tde_xlog_smgr.c
```diff
@@ -247,22 +248,37 @@ TDEXLogSmgrInitWrite(bool encrypt_xlog, int key_len)
...
-	{
-		pg_tde_create_wal_range(&CurrentWalEncryptionRange, WAL_ENCRYPTION_RANGE_ENCRYPTED, key_len);
-	}
-	else if (range && range->type == WAL_ENCRYPTION_RANGE_ENCRYPTED)
+#ifndef FRONTEND
+	if (IsUnderPostmaster)
 	{
-		pg_tde_create_wal_range(&CurrentWalEncryptionRange, WAL_ENCRYPTION_RANGE_UNENCRYPTED, key_len);
+		if (range)
+			CurrentWalEncryptionRange = *range;
```
jeltz, line 259: Why not just always assign it? Why treat NULL specially?

jeltz (changes requested):
Looks generally good but had some minor comments.

src/access/pg_tde_xlog_smgr.c
dutow: Because we dereference it, it can be null.

ci_scripts/build.sh
dutow: we do not use pg clfags, so this was necessary. I moved it into meson.build.

ci_scripts/build.sh
dutow: done

src/access/pg_tde_xlog_smgr.c
jeltz: Oh, I just misread the code.

jeltz (approved):

PR #623, opened 2026-06-15
PG-2473: MSVC portability fixes

dutow:
These changes fix most portability issues with MSVC.

One remaining issue after these is the KMIP library, which doesn't compile with it. We'll solve that by switching to the new library.

Not adding CI yet because it wouldn't work without the KMIP rework and EXEC_BACKEND fixes.

There are also additional PRs in the postgres repo that add a required DLLIMPORT attribute to md.


jeltz (approved):
The using inline functions instead of macros is just an improvement.

PR #624, opened 2026-06-17
PG-2470: Disable upstream smgr API with pg_tde

dutow:


jeltz (approved):

src/smgr/pg_tde_smgr.c
```diff
@@ -635,6 +635,10 @@ RegisterStorageMgr(void)
 	OurSMgrId = smgr_register(&tde_smgr, sizeof(TDESMgrRelation));
 	storage_manager_id = OurSMgrId;
 
+#if defined(PERCONA_API_VERSION) && PERCONA_API_VERSION >= 3
```
jeltz: Minor comment: can't we count on this always beign defined?

PR #625, opened 2026-06-19
Convert vault_v2 test into a TAP test

jeltz:
This allows us the assign a random port and skip the test on machines where OpenBao is not installed. The test case could certainly use more clenaup like improving the assertsions for the error cases or potentially splitting out the OpenBao code into a Perl module, but I think this is enough improvement to be worthwhile on its own.


t/vault_v2.pl
```diff
@@ -0,0 +1,228 @@
...
+		'kv');
+
+	# Create a test namespace for the tests to test namespace support
+	system_or_bail($bao_bin, 'namespace', 'create', 'pgns');
+	system_or_bail($bao_bin, 'secrets', 'enable', '-ns=pgns', '-path=secret',
+		'-description="Production Secrets"', 'kv-v2');
+
+	write_file("$temp/token", $cluster_info->{root_token});
+
+	{
```
dutow: This is just stylistic, but a return here would be nice?

t/vault_v2.pl
```diff
@@ -0,0 +1,228 @@
...
+
+	CREATE TABLE test_enc_ns (x int PRIMARY KEY) USING tde_heap;
+	INSERT INTO test_enc_ns (x) VALUES (1), (2);
+));
+
+$stdout = $node->safe_psql('postgres', 'SELECT * FROM test_enc_ns;');
+is($stdout, "1\n2", 'can read test_enc_ns');
+
+$stdout = $node->safe_psql('postgres', 'SELECT pg_tde_verify_key();');
+is($stdout, '', 'key verification succeeds with namesapces');
```
dutow: typo: namesapces

t/vault_v2.pl
```diff
@@ -0,0 +1,228 @@
...
+
+	wait_for_file("$temp/info", '.');
+
+	my $cluster_info = decode_json(slurp_file("$temp/info"));
+
+	local %ENV = %ENV;
+	$ENV{'VAULT_ADDR'} = "https://127.0.0.1:$port";
+	$ENV{'VAULT_CACERT'} = $cluster_info->{ca_cert_path};
+
+	# We need to enable key/value version 1 engine for just for tests
```
dutow: for the tests?

dutow (approved):

t/vault_v2.pl
jeltz: Fiuxed

t/vault_v2.pl
jeltz: Fixed

t/vault_v2.pl
jeltz: Fixed

PR #627, opened 2026-06-22
Fix pg_tde_rewind for encrypted data

dAdAbird:
pg_tde_rewind was totally broken with encrypted relations. The main reason is that it may mix source and target data for the same relation. Meaning the resulting relation would have data encrypted with different keys. To fix that, we re-encrypt incoming changes with the destination keys. But since the rest of the files that were copied from the source are encrypted with the source keys, we copy the source keys to the destination, but swap keys for the partially rewritten data to the proper one. There are numerous other changes and fixes:
- Some relations might stay unchanged on the destination, but since we copy source keys, we have to update keys for such relations as well.
- Forks share keys with the main fork. Meaning, any changes to the relation keys require forks re-encryption.
- Although usually WAL would be replaced by that of the source, sometimes rewind may keep some destination segments for recovery. Hence, those kept segments should be reencrypted with proper keys.
- Proper handling of wal restored from archive during rewind.
- pg_tde_rewind also exposed a couple of other issues, not related to the rewind per se.

See commit messages.

Fixes: PG-2397, PG-2234, PG-2358, PG-2407, PG-2363


dutow:
Looks good to me, just don't forget to remove the TMP commit.

jeltz (approved):

PR #630, opened 2026-06-26
Add missing SSL header for openssl 1.1

dutow:


Naeem-Akhter (approved):
Fix works fine on all platforms, verified it in pipelines.

PR #633, opened 2026-07-02
PG-2494 perf: vectorize WAL stream XOR

dutow:
The byte-at-a-time XOR in pg_tde_stream_crypt did not vectorize: the function supports in-place use (out == data), so without an aliasing guarantee the compiler cannot prove out and data are distinct.

Callers only ever pass fully overlapping (in-place) or fully disjoint buffers, never partial overlap, so select the XOR by a pointer equality check: in-place uses a plain loop, the disjoint case a restrict-qualified loop the compiler auto-vectorizes.

### DIFF
--- a/src/encryption/enc_tde.c
+++ b/src/encryption/enc_tde.c
@@ -69,7 +69,31 @@ pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
 }
 
 /*
- * Encrypts/decrypts `data` with a given `key`. The result is written to `out`.
+ * XOR n bytes of data with keystream into out.
+ *
+ * The distinct variant's restrict qualifiers let the compiler
+ * auto-vectorize the XOR, but require non-overlapping buffers.
+ * The plain variant is used for in-place operation (out == data).
+ */
+static inline void
+tde_xor_stream(char *out, const char *data, const unsigned char *keystream, uint32 n)
+{
+	for (uint32 i = 0; i < n; i++)
+		out[i] = data[i] ^ keystream[i];
+}
+
+static inline void
+tde_xor_stream_distinct(char *restrict out, const char *restrict data,
+						const unsigned char *restrict keystream, uint32 n)
+{
+	Assert(((uintptr_t) out + n <= (uintptr_t) data) || ((uintptr_t) out > (uintptr_t) data + n));
+	for (uint32 i = 0; i < n; i++)
+		out[i] = data[i] ^ keystream[i];
+}
+
+/*
+ * Encrypts/decrypts data with a given key. The result is written to out,
+ * which may alias data (in-place operation is supported).
  *
  * start_offset: is the absolute location of start of data in the file.
  */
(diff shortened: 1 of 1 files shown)


src/encryption/enc_tde.c
```diff
@@ -89,11 +112,18 @@ pg_tde_stream_crypt(const char *iv_prefix,
 	uint32		batch_no = 0;
 	uint32		data_index = 0;
 
+	/*
+	 * Our callers pass either fully overlapping (out == data, in-place) or
+	 * fully disjoint buffers, never partial overlap.
+	 */
+	const bool	in_place = (out == data);
```
jeltz, line 120: Isnt const here a C++-ism the PG codestyle does not use?

src/encryption/enc_tde.c
```diff
@@ -69,7 +69,30 @@ pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
 }
 
 /*
- * Encrypts/decrypts `data` with a given `key`. The result is written to `out`.
+ * XOR `n` bytes of `data` with `keystream` into `out`.
+ *
+ * The `distinct` variant's `restrict` qualifiers let the compiler
+ * auto-vectorize the XOR, but require non-overlapping buffers.
+ * The plain variant is used for in-place operation (`out` == `data`).
```
jeltz: We do not use markdown in comments,

src/encryption/enc_tde.c
```diff
@@ -69,7 +69,30 @@ pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
...
+tde_xor_stream_distinct(char *restrict out, const char *restrict data,
+						const unsigned char *restrict keystream, uint32 n)
+{
+	for (uint32 i = 0; i < n; i++)
+		out[i] = data[i] ^ keystream[i];
+}
+
+/*
+ * Encrypts/decrypts `data` with a given `key`. The result is written to `out`,
+ * which may alias `data` (in-place operation is supported).
```
jeltz: Shouldnät we add an assert to make sure there are no partial overlaps?

src/encryption/enc_tde.c
```diff
@@ -69,7 +69,30 @@ pg_tde_generate_internal_key(InternalKey *int_key, int key_len)
...
+static inline void
+tde_xor_stream_distinct(char *restrict out, const char *restrict data,
+						const unsigned char *restrict keystream, uint32 n)
+{
+	for (uint32 i = 0; i < n; i++)
+		out[i] = data[i] ^ keystream[i];
+}
+
+/*
+ * Encrypts/decrypts `data` with a given `key`. The result is written to `out`,
```
jeltz: More markdown in comments.

jeltz (changes requested):
The patch looks good but there are some stylistic issues plus that we should have an assert against partial overlaps.

src/encryption/enc_tde.c
dutow: I prefer to use it whenit makes sense even in C. The PG codestyle with all variables at the beginning make this somewhat harder, but not impossible. 3 of the 5 other local variables are const above

src/encryption/enc_tde.c
jeltz: Just remembered that comapring pointers to different objects i undefined so not sure you can assert against overlap. :D

dutow: It should work in practice, or I could cast them to uintptr_t

src/encryption/enc_tde.c
dutow: we should :) all ` removed

src/encryption/enc_tde.c
jeltz: Yeah, if we cast them it should not be undefined behviour. This is such a weird rule. I wonder what it allows.

src/encryption/enc_tde.c
jeltz: I think maybe we should stop using the PG code style but if so I would want a discussion in the team so we are on the same page.

jeltz (approved):
Do what you want with the remaining comments.

PR #634, opened 2026-07-06
PG-2125: Move libkmip pointer to the latest old-library commit (plan B)

dutow:
Old library was deleted after this commit.
Contains all fixes up to this point, including the vault workaround.


jeltz (approved):
Approved, but see my comment on Slack about way forward.

PR #635, opened 2026-07-06
PG-2536 PostgreSQL 19 support

jeltz:
Add support for PostgreSQL 19 Beta 2.

´Changes:

- Add frontend tools for PG 19
- Support that pg_rewind now can skip copying WAL files before the point of divergence
- Support splitting and merging partitions
- Meson file can now build against beta and rc versions
- Change the code for calculating AIO handles
- Some minor API changes
- Run pgindent with the version from PostgreSQL 19
- Add PG 19 to the CI
- Skip some more PSP tests which do not work with pg_tde


src/pg_tde_event_capture.c
```diff
@@ -252,6 +252,26 @@ alter_table_encryption_mix(Oid relid)
 	return enc;
 }
 
+static EncryptionMix
+alter_table_encryption_mix_check(Oid relid)
+{
+	EncryptionMix encmix = alter_table_encryption_mix(relid);
+
+	/*
+	 * This check is very braod and could be limited only to commands which
```
dAdAbird, line 261:
```suggestion
	 * This check is very broad and could be limited only to commands which
```

dAdAbird (commented):
Also, re [Patch frontend tools for PostgreSQL 19](https://github.com/percona/pg_tde/pull/635/changes/d0b39bb306ee9a0d7b14f46c464d35032a67818e):
1. Should we add a TAP test for this new "copy WAL" feature in pg_rewind?
2. Syntax typo in the commit message: "tries avoids" -> "tries to avoid". And if we're here, "where it tries" sounds better as "which tries"

src/pg_tde_event_capture.c
jeltz: Fixed!

dAdAbird (approved):

PR #641, opened 2026-07-27
PG-2588 Disable timeout in KMIP code as a workaround

jeltz:
The timeout code in libkmip is broken when used in PostgreSQL due to it using `SO_RCVTIMEO` which does not work well when combined with PostgreSQL which relies on `SA_RESTART` and signals.

But as we did not have timeout support in pg_tde 2.2.1 this removal does not affect our users.

### DIFF
--- a/src/keyring/keyring_kmip.cpp
+++ b/src/keyring/keyring_kmip.cpp
@@ -25,8 +25,11 @@ extern "C"
 
 namespace
 {
-/* Timeout applied to connect, handshake and every read/write, in ms. */
-constexpr int KMIP_TIMEOUT_MS = 10000;
+/*
+ * TODO: Timeouts are disabled since they are broken in libkmip due a bug with
+ *       SA_RESTART and signals.
+ */
+constexpr int KMIP_TIMEOUT_MS = 0;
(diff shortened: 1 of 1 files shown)


src/keyring/keyring_kmip.cpp
```diff
@@ -25,8 +25,11 @@ extern "C"
 
 namespace
 {
-/* Timeout applied to connect, handshake and every read/write, in ms. */
-constexpr int KMIP_TIMEOUT_MS = 10000;
+/*
+ * TODO: Timeouts are disabled since they are broken in libkmip due a bug with
+ *       SA_RESTART and signals.
+ */
+constexpr int KMIP_TIMEOUT_MS = 0;
```
artemgavrilov, line 32: Minor: It seems that we still can use real timeout for frontend part

artemgavrilov (approved):

src/keyring/keyring_kmip.cpp
jeltz: Yes, but I do not think it is worth it. Better to just properly fix libkmip.

PR #643, opened 2026-07-28
Do not reuse the curl handle across a fork

dutow:
The postmaster does keyring requests while generating the WAL key, so keyringCurl and its keepalive socket get inherited by every child. Two processes then interleave HTTP on the same connection, which surfaces as CURLE_WEIRD_SERVER_REPLY, a timeout, or a garbled response body.

Re-init the handle when the pid changed. The inherited handle is abandoned instead of cleaned up, because curl_easy_cleanup would write a TLS shutdown to the connection the parent still uses.

### DIFF
--- a/documentation/docs/release-notes/release-notes-v2.2.2.md
+++ b/documentation/docs/release-notes/release-notes-v2.2.2.md
@@ -38,3 +38,4 @@ Changes introduced in `pg_tde` 2.2.2:
 ### Bug Fixes
 
 - [PG-2492](https://perconadev.atlassian.net/browse/PG-2492) - Fixed crash when empty certificate parameters are passed to `pg_tde_add_global_key_provider_kmip()` or `pg_tde_add_database_key_provider_kmip()`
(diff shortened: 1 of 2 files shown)


jeltz (approved):
Looks good, but maybe we should merge it into `release-2.2` and update our release notes.

dutow:
@jeltz updated

documentation/docs/release-notes/release-notes-v2.2.2.md
```diff
@@ -38,3 +38,4 @@ Changes introduced in `pg_tde` 2.2.2:
 ### Bug Fixes
 
 - [PG-2492](https://perconadev.atlassian.net/browse/PG-2492) - Fixed crash when empty certificate parameters are passed to `pg_tde_add_global_key_provider_kmip()` or `pg_tde_add_database_key_provider_kmip()`
+- [PG-2608](https://perconadev.atlassian.net/browse/PG-2608) - Fix race condition in Vault key provider
```
Andriciuc: Slightly more descriptive: 

Fixed a race condition in the Vault key provider that could occur when multiple processes accessed the same cURL handle after a fork.

LGTM!

jeltz:
Looks good to me.

PR #644, opened 2026-07-28
PG-2407: Remove pg_rewind/pg_tde_rewind limitation, fixed in 2.2.1

ImTheKai:


janwieremjewicz (approved):
LGTM, the notice about pg_rewind is removed!

PR #645, opened 2026-07-30
Make WAL archive tools work with pg_wal as symlink

dAdAbird:
When deriving the pg_tde path from the WAL source, archive tools in the default env will get "pg_wal/../pg_tde". But it won't work if pg_wal is a symlink directing outside the datadir. So we add canonicalize_path to make such a path just "pg_tde".

Fixes PG-2609

### DIFF
--- a/t/wal_archiving_symlink.pl
+++ b/t/wal_archiving_symlink.pl
@@ -1,0 +1,89 @@
+# Check that WAL archiving works when pg_wal is a symlink.
+use strict;
+use warnings FATAL => 'all';
+use File::Copy;
+use PostgreSQL::Test::Cluster;
+use PostgreSQL::Test::Utils;
+use Test::More;
+
+# Test CLI tools directly
+
+# Wal archiving uses linux specific tmpfs mount point for temporary files.
+if ($^O ne 'linux')
+{
+	plan skip_all => 'tde wal_archiving works only on Linux';
+	exit 0;
+}
+
+my $keydir = PostgreSQL::Test::Utils::tempdir;
+my $wal_dir = PostgreSQL::Test::Utils::tempdir;
+
+# Test archive_command
+
+my $primary = PostgreSQL::Test::Cluster->new('primary');
+my $archive_dir = $primary->archive_dir;
+$primary->init(allows_streaming => 1);
+$primary->append_conf('postgresql.conf',
+	"shared_preload_libraries = 'pg_tde'");
+$primary->append_conf('postgresql.conf', "wal_level = 'replica'");
+$primary->append_conf('postgresql.conf', "autovacuum = off");
+$primary->append_conf('postgresql.conf', "checkpoint_timeout = 1h");
+$primary->append_conf('postgresql.conf', "archive_mode = on");
+$primary->append_conf('postgresql.conf',
+	"archive_command = 'pg_tde_archive_decrypt %f %p \"cp %%p $archive_dir/%%f\"'"
+);
+
+my $test_primary_datadir = $primary->data_dir;
+
+# Turn pg_wal into a symlink
+print("moving $test_primary_datadir/pg_wal to $wal_dir\n");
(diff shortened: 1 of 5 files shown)


t/wal_archiving_symlink.pl
```diff
@@ -0,0 +1,171 @@
...
+	`strings $data_dir/pg_wal/0000000100000000000000??`,
+	qr/foobar_enc/,
+	'should find foobar_enc in WAL since it is not encrypted');
+
+$replica_plain->promote;
+
+$result = $replica_plain->safe_psql('postgres',
+	'SELECT * FROM t1 UNION ALL SELECT * FROM t2');
+
+is($result, "foobar_plain\nfoobar_enc", 'b');
```
jeltz: The test should have a description.

t/wal_archiving_symlink.pl
```diff
@@ -0,0 +1,171 @@
...
+	`strings $data_dir/pg_wal/0000000100000000000000??`,
+	qr/foobar_enc/,
+	'should not find foobar_enc in WAL since it is encrypted');
+
+$replica_enc->promote;
+
+my $result = $replica_enc->safe_psql('postgres',
+	'SELECT * FROM t1 UNION ALL SELECT * FROM t2');
+
+is($result, "foobar_plain\nfoobar_enc", 'b');
```
jeltz: The test should have a description.

jeltz (changes requested):

jeltz:
You also need to update the changelog/release notes.

dAdAbird:
- I cleaned up the test and left only the bare minimum for this case. Before, it was mainly a copy of `wal_archiving.pl`.
- Added CHANGELOG

t/wal_archiving_symlink.pl
```diff
@@ -0,0 +1,88 @@
...
+$replica->start;
+
+$replica->wait_for_log("waiting for WAL to become available");
+
+$replica->promote;
+
+# Check that data restored from the WAL archive is readable
+my $result = $replica->safe_psql('postgres', 'SELECT * FROM t1');
+
+is($result, "foobar", 'b');
```
jeltz: Seems like you missed this one.

t/wal_archiving_symlink.pl
```diff
@@ -0,0 +1,88 @@
...
+$primary->append_conf('postgresql.conf',
+	"archive_command = 'pg_tde_archive_decrypt %f %p \"cp %%p $archive_dir/%%f\"'"
+);
+
+my $test_primary_datadir = $primary->data_dir;
+
+# Turn pg_wal into a symlink
+print("moving $test_primary_datadir/pg_wal to $wal_dir\n");
+move("$test_primary_datadir/pg_wal", $wal_dir) or die;
+dir_symlink($wal_dir, "$test_primary_datadir/pg_wal") or die;
```
jeltz: Use `BAIL_OUT()` instead of `die()` here so you can provide an error message.

jeltz (changes requested):
Is the idea to merge this into `release-2.2`? If so we should likely update the release notes too.

t/wal_archiving_symlink.pl
dAdAbird: done

t/wal_archiving_symlink.pl
dAdAbird: I didn't quite get what you mean.

dAdAbird:
Added release notes

t/wal_archiving_symlink.pl
jeltz: I mean `'b'`.

jeltz:
I think you have to change the target branch to `release-2.2`.

t/wal_archiving_symlink.pl
dAdAbird: Ah, yesss

dAdAbird:
all done

jeltz (approved):

PR #646, opened 2026-08-03
PG-2613: Document WAL encryption backup requirements

ImTheKai:


documentation/docs/command-line-tools/pg-tde-archive-decrypt.md
```diff
@@ -53,8 +53,16 @@ archive_command='pg_tde_archive_decrypt %f %p "cp %%p /mnt/server/archivedir/%%f
...
+
+    You must also disable the following pgBackRest features in the applicable pgBackRest configuration section:
+
+    ```ini
+    archive-async=n
+    archive-header-check=n
+    checksum-page=n
+    ```
+
+    You must also disable PostgreSQL data checksums. Otherwise, pgBackRest may emit excessive error messages, and in some package builds the log statement can cause crashes.
```
Andriciuc: You must also disable PostgreSQL data checksums, and the following pgBackRest features in the applicable pgBackRest configuration section:

    ```ini
    archive-async=n
    archive-header-check=n
    checksum-page=n
    ```

Otherwise, pgBackRest may emit excessive error messages, and in some package builds the log statement can cause crashes.

Andriciuc: same for below 🙏

ImTheKai: If I change the order, this reads like the config setting below disabled postgres checksums, which it doesn't, as this is only for pgBackRest

Andriciuc: Hmm, thank you! Ok then keep the text as is but the first line change to:

Disable the following pgBackRest features...
```
...
```

You must also disable...

So we avoid the repetition. 🙏

ImTheKai: Good point, thanks, fixed.

jeltz (approved):

